PR #5, opened 2015-10-02
First stab at Rust support for charmtools

cholcombe973:
Just like it says.  This is my first attempt support for doing a charm-create -t rust 
I'd love some feedback because I was not entirely sure that what I did would be obvious to end users.


chuckbutler:
@marcoceppi - we can probably close this PR, as its been stated we're going to focus on interpreted languages for charms, whereas rust would be a direct antipattern to that objective.

PR #17, opened 2015-10-14
Move LPGitFetcher upstream and add missing call to rename()

### LINKED ISSUE #14: Git repos on Launchpad are not handled properly
In `charmtools/fetchers.py` the test for a bzr repo matches anything on Launchpad, and the test for Git(Hub) repos only matches GitHub.  This causes git repos on Launchpad to be fetched with bzr, which works, but seems to cause subtle differences, or may not always work.

johnsca:
Fixes #14

### DIFF
--- a/charmtools/compose/fetchers.py
+++ b/charmtools/compose/fetchers.py
@@ -1,5 +1,3 @@
-import re
-import tempfile
 import os
 
 import requests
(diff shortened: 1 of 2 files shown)


marcoceppi:
LGTM

PR #21, opened 2015-10-20
Build name

bcsaller:
This renames generate/compose to build
It changes composer.yaml to layer.yaml (with some backwards compat)
COMPOSER_PATH -> LAYER_PATH


marcoceppi:
This is great, thanks for taking time to knock this out. LGTM :boat:

PR #45, opened 2015-11-10
Implement charm-proof for storage

### LINKED ISSUE #41: proof doesn't know about storage
```
INFO:root:Workspace dir: /var/lib/jenkins/workspace/charm_lint_check
INFO:root:Reading file: Makefile
INFO:root:Searching for: ['@flake8', '/flake8']
INFO:root:command:  make -f Makefile lint
E: Unknown root metadata field (storage)
I: config.yaml: option ephemeral-unmount has no default value
I: config.yaml: option monitor-secret has no default value
I: config.yaml: option ceph-public-network has no default value
I: config.yaml: option ceph-cluster-network has no default value
I: config.yaml: option key has no default value
I: config.yaml: option fsid has no default value
I: config.yaml: option osd-journal has no default value
I: config.yaml: option osd-reformat has no default value
I: config.yaml: option source has no default value
make: *** [lint] Error 200
ERROR:root:Make target returned non-zero.
```

It should probably be pretty similar to the relation checking.

tvansteenburgh:
Added colander dep for schema validation. Already packaged for apt
so this shouldn't add any packaging burden down the road.

There were a handful of +x files in the tests/ dir, which was
(curiously) causing nosetests to not find those tests. After setting
-x, nose detected about 50 additional tests that weren't being run
before. Many of those (in test_juju_test.py) were failing because
the cwd was not being set properly to allow the Conductor to find
the tests. So that's what all those updates are for...

The change to record.sh is to ensure the tests run successfully
after a clean repo checkout (they failed for me originally).

(Fixes #41)


axw:
validate_storage LGTM, thank you! I think someone else ought to review the other bits.

marcoceppi:
LGTM +1 :boat:

PR #48, opened 2015-11-17
Added support for building a wheelhouse

johnsca:
Each layer can now have a `wheelhouse.txt` requirements file that will be built into a `--no-binary` (source-only) wheelhouse.


chuckbutler:
WHAT? :beers:

johnsca:
This depends on a newer version of pip than is available in Trusty.  Having charm-tools depend on the newer version of pip would likely cause issues with packaging for apt.  @marcoceppi suggested creating a venv and installing the newer pip in that.  Does that seem  reasonable?

marcoceppi:
@johnsca What version of pip is needed? If it's in Xenial it'll be easy to backport to trusty and it'll be fine for when we upload charm-tools to Xenial. If it's not in Xenial that's still OK we just have a little extra work to put forth there.

johnsca:
It requires version 7.1.0.  There was a version numbering change from 1.5.6 -> 6.0 (the leading 1 was dropped), and Trusty has 1.5.4 from Feb 2014.

stub42:
Check with Barry before backporting pip - I think there might be Ubuntu specific tweaks needed to make it play nicely with apt packages (and Barry might already have packages somewhere). A virtualenv or using fades might be simpler. Xenial currently has 1.5.6, but that should change before release.

(heck, since we are already pulling down packages from pypi we could use old-pip to install new-pip in a temporary directory then use new-pip to install the wheels into the wheelhouse.)

johnsca:
@stub42 We did get in touch with Barry and he was already planning to get the pip package updated.  We're going to assist with that work and then get backported to the juju/stable trusty, et al. PPAs.

I'm ok with the idea of building a pip venv to then use to build the wheelhouse, but we are hopeful that we can get the backporting work done within the next iteration or so.  If it  ends up taking longer, or if I get impatient, I might do the venv approach in the interim.

johnsca:
Ok, switched to use a virtualenv until such time that python-pip is more up-to-date.  It also includes that newer pip as a wheel in the wheelhouse, so that the charm can use it to bootstrap the wheelhouse (otherwise it gets spurious "double requirements" errors when installing the wheelhouse at deployment time).

I also fixed the file tracking so that it now associates all files with a layer and we should no longer end up having to use `--force` on rebuilds.

marcoceppi:
LGTM :boat:

Issue #44, opened 2015-11-07
Relative paths don't work with charm build in the output flag.


kat-co:
[This](https://github.com/kat-co/elnode-charm/blob/master/Makefile#L4) is effectively using a relative path to direct output to a distribution folder. Through working with @bcsaller @marcoceppi @johnsca and @mbruzek, we discovered that this fails to produce a bin directory in the generated charm.

@bcsaller has asked that I assign this to him.

johnsca:
This is quite possibly resolved with #54.

PR #54, opened 2015-11-25
Fixed issues with relative paths in charm build

johnsca:


marcoceppi:
Does this fix #44 ?

johnsca:
I believe it should, but I didn't do a full audit of relative path issues.

marcoceppi:
:boat:

PR #55, opened 2015-12-01
Use Python 3 for layered charms

johnsca:
This depends on juju-solutions/reactive-base-layer#10.


marcoceppi:
Does this mean python3-pip and python3-dev are required deps for charm-tools?

johnsca:
It would depend on python3 and python3-pip until we remove the InstallerTactic.  The WheelhouseTactic uses virtualenv, so I'm not sure if that requires python3-pip or not.

I can't see why it would depend on python3-dev

PR #62, opened 2015-12-03
Fix wheelhouse containing platform specific wheels

johnsca:
Apparently, I completely misunderstood the point of the `--no-binary` option for `pip wheel`.  It actually controls what `pip` pulls down from pypi.python.org and not what gets generated in the wheelhouse.  I've been told that wheels are inherently binary, pre-compiled, and platform specific.

But it seems that using `pip install --download wheelhouse/ --no-binary :all:` instead of `pip wheel -w wheelhouse/ --no-binary :all:` gets us what we want.  Technically, the result is no longer a wheelhouse, since it contains (only) sdists, so I'd be ok with changing the name to "pydeps" and "pydeps.txt" or similar, but for now I left it as "wheelhouse".

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -493,9 +493,9 @@ class WheelhouseTactic(ExactMatch, Tactic):
     def _add(self, pip, wheelhouse, *reqs):
         with utils.tempdir(chdir=False) as temp_dir:
             # put in a temp dir first to ensure we track all of the files
-            utils.Process((pip, 'wheel',
+            utils.Process((pip, 'install',
                            '--no-binary', ':all:',
-                           '-w', temp_dir) +
+                           '-d', temp_dir) +
                           reqs).throw_on_error()()
             for wheel in temp_dir.files():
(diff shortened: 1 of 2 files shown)


johnsca:
We will want to merge juju-solutions/reactive-base-layer#11 at the same time.

marcoceppi:
:boat:

wwitzel3:
They will only be platform specific if they have C extensions. As is the
case for PyYAML.

johnsca:
@wwitzel3 Yes, but some of the deps coming from the base layer include optional C extensions (with fallbacks to pure Python), so the wheels fail to install.  Ideally, we'd have a way to force it to always create pure Python wheels if at all possible and fail otherwise and then just assert that you must use libraries with pure Python fallbacks or manage them yourself (since you'll have to do extra work to ensure that they build), but that doesn't seem to be something that can be done.  So, we're left in a situation where we use the sdist from pypi and there will be no build-time reporting that a given dep can't be managed by the base layer.

PR #64, opened 2015-12-04
Add dist.yaml and resources.yaml as merged YAML files

johnsca:
These are blockers for the big data layered charms.


marcoceppi:
:sail:

PR #68, opened 2015-12-05
Resolve #65: Add support for layer options in layer.yaml

### LINKED ISSUE #65: Add support for layer options in layer.yaml
It is useful for base layers to accept some manner of "config" from layers building on them and YAML files are a reasonable way to accomplish this.  However, if each layer defines its own YAML file (e.g., the [Apache PHP base layer](https://github.com/juju-solutions/layer-apache-php) reads an `apache.yaml` file), this will lead to a proliferation of YAML files and increase the potential for collision.  It also doesn't cleanly support subsequent layers override the layer config of a previous layer, without having to replicate the entire YAML file.

The `layer.yaml` file should support two new sections: `options` and `defines`*.

The `defines` section should contain a [jsonschema](https://pypi.python.org/pypi/jsonschema) definition describing what options it supports and what their values can be.  (This is the same format that [actions](https://jujucharms.com/docs/stable/authors-charm-actions) use to define their params.)  These options will automatically be namespaced by the layer name.

The `options` section will be a dictionary section with layer names as keys, and the set of options & values for that layer, as defined by that layer's `defines` section.  The `charm build` command can use the schemas from the `defines` section to validate the options.  The `layer.yaml` file of a higher level layer can override the options specified by lower level layers, although obviously care should be taken if doing so.

See https://gist.github.com/marcoceppi/ad318d081e3b30857819 for an example of what this would look like.

*: Feel free to bike-shed on the names.

johnsca:
For #65.  Add support for jsonschema defined and validated options in ``layer.yaml`` for configuring included layers.


charmtools/build/tactics.py
```diff
@@ -353,11 +387,34 @@ def trigger(cls, relpath):
...
+
+    def combine(self, existing):
+        super(LayerYAML, self).combine(existing)
+        self.schema = utils.deepmerge(existing.schema, self.schema)
+        return self
+
+    def lint(self):
+        if 'options' not in self.data:
+            return True
+        validator = extend_with_default(jsonschema.Draft4Validator)(self.schema)
```
bcsaller: It looks like in this case 'lint' is preforming a now needed mutation of the object which is a bit odd conceptually in that other tactics would be expected to do something similar. Can we do this in 'read' which should be related to loading the version of that data we expect to use in processing (maybe read should be 'load' or 'process' or 'parse' but that is another papercut).

johnsca: @bcsaller I agree that it's not ideal to do modification of the target values during validation, but that's unfortunately what the jsonschema library [recommended](https://python-jsonschema.readthedocs.org/en/latest/faq/#why-doesn-t-my-schema-that-has-a-default-property-actually-set-the-default-on-my-instance).  Ideally, the default setting behavior would be handled just after `combine` but I'm not clear on how to implement that without a lot of manual processing of the schema.

PR #72, opened 2015-12-07
add /home/marco to whitelist fixes #47

### LINKED ISSUE #47: $HOME is needed for juju-test with LXD provider
Removing the HOME env var, the $HOME/.config dir becomes /.config and lxd-provider cannot locate lxd.

Workaround: run juju-test with -p HOME

Expected behavior: workaround should be unnecessary.

marcoceppi:


chuckbutler:
+1

Issue #74, opened 2015-12-08
Move hook template out to layer


johnsca:
https://github.com/juju/charm-tools/blob/master/charmtools/build/tactics.py#L14

This template should be provided by the layers and not embedded in charm-tools.  Define a standard for layers to provide it, and provide the current default in layer-basic.

Issue #70, opened 2015-12-07
Generated storage-attached hooks fail


axw:
I'm building a reactive charm that has storage, and it fails on storage-attached. I'm using charm-tools 1.10. The error is:
```
ImportError: No module named 'charms'
```

storage-attached may run before the install hook. It looks like the charms.reactive package doesn't get installed until the install hook.

PR #75, opened 2015-12-08
Use hook template from layers instead of embedding

johnsca:
Implement #74.  This will allow iteration on the hooks w/o release of charm-tools, as well as non-Python base layers.


charmtools/build/__init__.py
```diff
@@ -334,7 +337,30 @@ def plan_interfaces(self, layers, output_files, plan):
                 plan.append(
                     charmtools.build.tactics.InterfaceBind(
                         iface, relation_name, kind,
-                        self.target, target_config))
+                        self.target, target_config, template))
+
+    def plan_storage(self, layers, output_files, plan):
+        # Interface includes don't directly map to output files
```
bcsaller: copy-paste error on comment

charmtools/build/__init__.py
```diff
@@ -119,6 +115,7 @@ class Builder(object):
     Handle the processing of overrides, implements the policy of BuildConfig
     """
     PHASES = ['lint', 'read', 'call', 'sign', 'build']
+    HOOK_TEMPLATE_FILE = path('hooks/hook.template')
```
marcoceppi: It seems odd that we define this several times. Here, L12 & L202 of tactics.py, etc

charmtools/build/tactics.py
```diff
@@ -253,6 +235,40 @@ def __str__(self):
         return "Bind Interface {}".format(self.interface.name)
 
 
+class StorageBind(Tactic):
```
bcsaller: Should this share a baseclass with interface bind and share code. replicating things like sign() happen too much already in the code base. I won't insist, but if you agree that is a good idea we can make an issue.

charmtools/utils.py
```diff
@@ -520,3 +520,7 @@ def prefix_lines(lines, lineno):
                    m=message)
         i += 1
     return i == 0
+
+
+class BuildError(Exception):
```
bcsaller: maybe move this to an exceptions module, it isn't really a utility, just a nitpick though

johnsca: I didn't actually mean to move this since I didn't end up using it in tactics.py like I thought I might need to.

charmtools/build/__init__.py
johnsca: That was a mistake.  Addressed.

charmtools/build/tactics.py
```diff
@@ -215,42 +196,53 @@ def lint(self):
...
                          "dynamic",
                          utils.sign(target))
         return sigs
 
     def __str__(self):
-        return "Bind Interface {}".format(self.interface.name)
+        return "{}: {}".format(self.__class__.__name__, self.name)
+
+
+class InterfaceBind(DynamicHookBind):
```
bcsaller: Thank you :)

chuckbutler:
From what i see here wrt the storage hook generators, this looks +1 to me, however i haven't verified by attempting to build a charm w/ storage hooks using this branch. YMMV w/ that +1

axw:
> From what i see here wrt the storage hook generators, this looks +1 to me, however i haven't verified by attempting to build a charm w/ storage hooks using this branch. YMMV w/ that +1

I won't have time to test until next week, but if someone wants to test storage, just add the following to a charm's metadata.yaml:
```
storage:
    whatever:
        type: filesystem
```

This will create a required filesystem store called "whatever". Deploying without storage constraints should create a directory in the root filesystem that will be used for that store. Just deploying a generated charm that has required storage currently fails, as noted in #70.

johnsca:
@axw This change is a pre-req for fixing #70.  To test this, you will need to update the basic layer to include the template.  I will get that proposed now to make this easier to test.

johnsca:
This can be tested with juju-solutions/reactive-base-layer#16.  That PR can be merged before this one.

PR #90, opened 2016-01-21
pip 8.0.0 wreaks havoc. lock it to 7.X

marcoceppi:
This either needs to be handled here, or pip should simply be defined in the "base" layer and L624 can just be removed

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -621,7 +621,6 @@ class WheelhouseTactic(ExactMatch, Tactic):
         if create_venv:
             utils.Process(('virtualenv', '--python', 'python3', venv)).throw_on_error()()
             utils.Process((pip, 'install', '-U', 'pip', 'wheel')).throw_on_error()()
-            self._add(pip, wheelhouse, 'pip')
         for tactic in self.previous:
             tactic(venv)
         self._add(pip, wheelhouse, '-r', self.entity)


bcsaller:
I'd rather track current pip. Let us not pin on error rather than fix. Looks like the default wheel-dir changed so we need to tell pip where to pull (from wheelhouse/ now that cwd is the default)

diff --git a/charmtools/build/tactics.py b/charmtools/build/tactics.py
index 980c27c..899f16a 100644
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -604,6 +604,7 @@ class WheelhouseTactic(ExactMatch, Tactic):
             # put in a temp dir first to ensure we track all of the files
             utils.Process((pip, 'install',
                            '--no-binary', ':all:',
+                           '-f', 'wheelhouse',
                            '-d', temp_dir) +
                           reqs).throw_on_error()()
             for wheel in temp_dir.files():

marcoceppi:
That's fine, but all charms being built are broken I'd rather patch now
then address later

johnsca:
@bcsaller Recapping what we discussed out-of-band, the issue isn't related to the `--find-links` changes in pip but, rather, in a conflict between the `python3-six` apt package and the updated dependency in pip 8.0.0.  And this PR is now just about moving the pip wheelhouse entry to the base layer where it better fits, and we can look to addressing the conflict there in the future.

PR #100, opened 2016-02-04
Added PowerShell biolerplate generation.

aznashwan:
This PR brings support for generation of a powershell template with `charm create`.

We chose to just do an exec call to git for a couple of reasons:
*  the boilerplate code itself depends on our [Juju Modules](https://github.com/cloudbase/juju-powershell-modules/tree/master).
* the Juju modules must remain a stand-alone project, as they are generally useful.
* the template's code and the Juju modules must stay synced.
* as the old Romanian saying "may the neighbour's goat die too" goes; the Python template's setup does an exec call to bzr too, so requiring git doesn't seem that unreasonable...

### DIFF
--- a/charmtools/create.py
+++ b/charmtools/create.py
@@ -86,6 +86,9 @@ def main():
             "Using default charm template (%s). To select a different "
             "template, use the -t option.", DEFAULT_TEMPLATE)
         args.template = DEFAULT_TEMPLATE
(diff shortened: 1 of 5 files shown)


marcoceppi:
YES! PowerShell templates! +1 LGTM :boat:

marcoceppi:
Will merge against road-to-2.0 branch

Issue #99, opened 2016-02-03
charm build on yaml needs to do proper type casting


marcoceppi:
When you have a `config.yaml` with the following:

```yaml
options:
  install_keys:
    type: string
    description: The key for the Debian package repository.
    default: "01182252"
```

Because this is actually a string but /may/ be a string of INTs Yaml parses this to an integer, writes the built YAML as:

```yaml
options:
  install_keys:
    type: string
    description: The key for the Debian package repository.
    default: 01182252
```

When writing the YAML back out, charm build should typecast based on `type` of the config options (string, int, bool)

To test, check out @srbaker [neo4j layer](https://github.com/srbaker/neo4j-charm)

johnsca:
It seems to be an issue in pyyaml, and is due to the leading zero.

mbruzek:
Actually @johnsca it looks like the tactics.py class imports ruamel to preserve the comments on a round trip. I think we need to focus on options we can pas ruamel or fix this problem there.

mbruzek:
This problem is due to the ruamel library that is use in the charm build python classes. After installing the latest version of ruamel (0.10.20), I can still reproduce this bug with strings that start with zero and are over nine characters in length.  

I created a bug against ruamel for this problem: https://bitbucket.org/ruamel/yaml/issues/23/strings-with-leading-zeros-are-round

Here is the sample code that reproduces this bug.
```python
#!/usr/bin/env python

from ruamel import yaml

yaml_string = """\
options:
  key: "{0}"
"""
value = ""
for x in range(0,12): 
    value = value + str(x)
    string = yaml_string.format(value)
    data = yaml.load(string, Loader=yaml.RoundTripLoader)
    print(yaml.dump(data, Dumper=yaml.RoundTripDumper))
```

stub42:
Accoding to upstream, the unquoted 01182252 is still a string. Its not a decimal because it starts with '0'. Its not an octal number because it contains an 8. So it falls back to being a string, and no quotes are needed.

pyyaml agrees with this:
```python
>>> type(yaml.safe_load("'09876534'"))
<type 'str'>
>>> type(yaml.safe_load("09876534"))
<type 'str'>
>>> type(yaml.safe_load("076534"))
<type 'int'>
```

marcoceppi:
Then this is a bug in go-yaml parsing

mbruzek:
[...]

PR #104, opened 2016-02-06
Addresses issue 99 where Juju has a problem with strings starting with zero

mbruzek:
This pull forces all strings to use quotes so Juju does not try to read the yaml string 01182252 as an int.

Adding configuration and tests so this does not revert in the future.

I also had to fix a hash on metadata.yaml in the tests, don't know what that was about other than the test was failing when I found it.

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -319,7 +319,8 @@ class YAMLTactic(SerializedTactic):
     def dump(self, data):
         yaml.dump(data, self.target_file.open('w'),
                   Dumper=yaml.RoundTripDumper,
-                  default_flow_style=False)
+                  default_flow_style=False,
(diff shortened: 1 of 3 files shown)


marcoceppi:
Lgtm, the hash is the has of the file. So this needs to be updated since content has changed.

Did not QA, will wait for travis

marcoceppi:
Fixes #99

mbruzek:
@marcoceppi the hash was for metadata.yaml not config.yaml, so I don't know why it changed.

marcoceppi:
@mbruzek can you open a bug against the go-yaml library Juju uses to fix this at its source?

mbruzek:
https://github.com/go-yaml/yaml/issues/157

PR #107, opened 2016-02-12
Fixed #105: Raise error if interface is missing implementation for role being used

### LINKED ISSUE #105: At build time, should warn user if they impement a relation which does not have code
There are several impartial implementations of an interface in the interface site.

If a user has the following:

```yaml
provides:
  relation:
    interface: you-knows-it
```

and `interface:you-knows-it` only has a `requires.py` then the user should be faced with a WARNING or potentially ERROR that the relation will not be implemented properly.

johnsca:
```
build: Composing into /home/johnsca/juju/envs/apache-dev
build: Processing layer: layer:basic
build: Processing layer: hadoop-client
build: Processing interface: hadoop-plugin
charmtools.build.tactics: Missing implementation for interface role: provides.py
```

### DIFF
--- a/charmtools/build/__init__.py
+++ b/charmtools/build/__init__.py
@@ -313,10 +313,10 @@ class Builder(object):
         target_config = layers["layers"][-1].config
         specs = []
         used_interfaces = set()
-        for kind in ("provides", "requires", "peers"):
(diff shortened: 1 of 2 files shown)


marcoceppi:
This is awesome! very helpful. :boat: +1

PR #108, opened 2016-02-12
Fixed #102: Report useful error message when fetch fails due to missing VCS tool

### LINKED ISSUE #102: charm build fails with a large stacktrace when git is not present on the system
when executing `charm build` in a fresh environment without installing git, the process fails with a long stack trace referring to 'file not found' - not indicating that user needs to install git. This seems to be a non issue on packaged versions of charm-tools, however when installing from pip, it wont implicitly pull in git.

johnsca:
```
build: Composing into /home/ubuntu/t
build: Unable to run "git": No such file or directory
```

### DIFF
--- a/charmtools/fetchers.py
+++ b/charmtools/fetchers.py
@@ -1,3 +1,4 @@
+import errno
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


charmtools/fetchers.py
```diff
@@ -308,12 +308,15 @@ class FetchError(Exception):
...
-    )
+    try:
+        p = subprocess.Popen(
+            args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            **kw
+        )
+    except OSError as e:
+        raise FetchError('Unable to run "%s": %s' % (args[0], e.strerror))
```
marcoceppi:
```python
msg = 'Unable to run "%s": %s' % (args[0], e.strerror)
if e.errno == 127:
    msg = msg + '\nPlease install "%s" before continuing.' % args[0]
raise FetchError(msg)
```

That way the user has an actionable item?

marcoceppi:
:boat: LGTM

PR #112, opened 2016-02-12
Fixed #97: Layer options defaults not set if options section missing

### LINKED ISSUE #97: Layer options defaults not set if options section missing
If no layer's `layer.yaml` includes an options section, no options end up in the rendered `layer.yaml` even if defaults are set.  Even having an empty options section (i.e., `options: {}`) is sufficient to get the options with defaults to render.

johnsca:


marcoceppi:
LGTM :boat:

PR #113, opened 2016-02-12
Fixed #77: Added templates for reactive charms

### LINKED ISSUE #77: `charm create` should output a reactive charm
When starting a new charm with `charm create`, the output should be a reactive charm. While it's certainly not hard to write a reactive charm from scratch (because it's so easy to write a reactive charm now!), I think it's important to update this, as it's possibly the first thing a learning charmer will try.

johnsca:
Set reactive-python to the default

### DIFF
--- a/charmtools/create.py
+++ b/charmtools/create.py
@@ -31,7 +31,7 @@ from charmtools.generators import (
 
 log = logging.getLogger(__name__)
 
-DEFAULT_TEMPLATE = 'python'
+DEFAULT_TEMPLATE = 'reactive-python'
 
 
 def setup_parser():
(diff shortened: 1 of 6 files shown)


marcoceppi:
Would it be better to host these in `github.com/juju-solutions/reactive-{python,bash}` and just have the create command clone them? That way as we iterate on the templates we don't have to push a release out?

Reference: https://github.com/juju/charm-tools/pull/100/

johnsca:
Agreed. I was trying to bang these out pretty quickly, but I will refactor it to follow that pattern.

marcoceppi:
LGTM :boat:

Issue #122, opened 2016-03-02
Implement repo as valid, recommended, key in layer.yaml


marcoceppi:
In `layer.yaml` we should strongly recommend that an additional `repo` key which will point to the upstream repository for that layer. When not present, build should WARN that this isn't set. `charm proof` should also WARN when a .build.manifest file is present but a repository isn't listed.

Example values:

- https://github.com/user/layer-repo
- git@github.com:user/layer-repo
- lp:layer-repo
- lp:~user/+junk/layer-repo
- https://git.launchpad.net/layer-repo
- bzr+ssh://bazaar.launchpad.net/layer-repo

All of these should be accessible from their respective tools directly as valid urls

PR #127, opened 2016-03-11
During `charm build`, suggest a 'repo' key for layer.yaml (if missing)

tvansteenburgh:
Prints a WARN message if repo key is not present. If the top layer is in a vcs directory (git, bzr, or hg), will also provide a suggested repo url.


marcoceppi:
LGTM :boat:

tvansteenburgh:
Fixes #122

PR #133, opened 2016-03-15
Avoid multiple fetches of a layer or interface during charm-build.

### LINKED ISSUE #53: multiple layers that inherit layer-basic gets downloaded multiple times
An example output of my ghost layer https://github.com/battlemidget/juju-charm-ghost which inherits a few layers which all inherit the base layer as their starting point:

http://paste.ubuntu.com/13503577/

Lines 19, 21, 23 show the build process cloning the base layer multiple times. This doesn't break anything but does add an additional wait time when building out the layers

tvansteenburgh:
Fixes #53

### DIFF
--- a/charmtools/build/__init__.py
+++ b/charmtools/build/__init__.py
@@ -256,17 +256,19 @@ class Builder(object):
             baselayers = [baselayers]
 
         for base in baselayers:
+            # The order of these commands is important. We only want to
+            # fetch something if we haven't already fetched it.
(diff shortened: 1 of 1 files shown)


tvansteenburgh:
Here's a before-and-after snip from the build log, using the ghost charm (https://github.com/battlemidget/juju-charm-ghost) cited in #53: http://pastebin.ubuntu.com/15392477/

johnsca:
:boat:

PR #134, opened 2016-03-15
Support layer- and interface- prefix for local layers

### LINKED ISSUE #116: Support layer- and interface- prefix for local layers
The convention that has come up for repo names for base, interface, and charm layers is to prefix them with ["layer-"](https://github.com/juju-solutions?query=layer-) and ["interface-"](https://github.com/juju-solutions?query=interface-).  When those repos are cloned for local development, you have to be careful to rename the directory to remove the prefix.  It would be nice if `charm build` would handle the prefix automatically.

tvansteenburgh:
When the fetcher searches for local layers and interfaces, look
for a prefixed dir as well as a dir that matches the layer or
interface name exactly. A convenience for users who have cloned
conventionally-named layer/interface repos into their
LAYER_/INTERFACE_PATH. Fixes #116.

### DIFF
--- a/charmtools/build/fetchers.py
+++ b/charmtools/build/fetchers.py
@@ -36,22 +36,24 @@ class InterfaceFetcher(fetchers.LocalFetcher):
 
     @classmethod
     def can_fetch(cls, url):
-        # Search local path first, then
-        # the interface webservice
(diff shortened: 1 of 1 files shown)


johnsca:
LGTM :boat:

PR #135, opened 2016-03-15
Add --show-report option to charm-build

johnsca:
This option enables a post-build report detailing what files have changed, based on the signatures in the existing `.build.manifest`

### DIFF
--- a/charmtools/build/__init__.py
+++ b/charmtools/build/__init__.py
@@ -174,6 +174,10 @@ class Builder(object):
     def charm_name(self):
         return "{}/{}".format(self.series, self.name)
 
+    @property
+    def manifest(self):
+        return self.target_dir / '.build.manifest'
+
     def status(self):
         result = {}
         result.update(vars(self))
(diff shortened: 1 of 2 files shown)


chuckbutler:
YES!

johnsca:
Making some refactors to use existing delta_signatures logic as suggested in IRC.

charmtools/build/__init__.py
```diff
@@ -588,6 +623,8 @@ def main(args=None):
                         default="http://interfaces.juju.solutions")
     parser.add_argument('-n', '--name',
                         help="Build a charm of 'name' from 'charm'")
+    parser.add_argument('-r', '--show-report', action="store_true",
```
marcoceppi: NIT: `--report` or `--with-report`?

charmtools/build/__init__.py
```diff
@@ -518,6 +520,39 @@ def clean_removed(self, signatures):
...
             filepath.remove()
 
+    def write_report(self, new_sigs):
+        """
+        Log messages indicating what changed with this (re)build.
+        """
+        old_sigs = self.old_manifest['signatures']
+        if not old_sigs:
+            log.info('')
+            log.info('---------------------------------------')
```
marcoceppi: Would we ever want report data to be machine readable?

johnsca:
Made some changes based on review.

@marcoceppi The report format is basically machine-readable but is still mixed in with the log output.  I suppose we might want to dump it to a file at some point with less header info, but at that point it could maybe just  be added to the `.build.manifest` file in a "changes" section or something.

marcoceppi:
LGTM :BOAT:

bcsaller:
LGTM, thanks

PR #148, opened 2016-03-22
Replace charmworldlib with libcharmstore

### LINKED ISSUE #144: Remove charmworldlib from charm-tools
replace with libcharmstore

tvansteenburgh:
- Removed all the remote proofing stuff. Now we just proof
  locally with jujubundlelib, with some of our own checks
  on top.
- Had to pin requests at 2.6.0 b/c theblues, a dep of
  libcharmstore, requires that specific version.
- Integration tests ('make check') all pass.

Fixes #144.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 PyYAML==3.11
 blessings==1.6
 bzr>=2.6.0
-libcharmstore>=0.0.2
+libcharmstore>=0.0.3
 cheetah==2.4.4
 colander==1.0
 coverage==3.7.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 PyYAML==3.11
 blessings==1.6
 bzr>=2.6.0
-charmworldlib>=0.4.2
+libcharmstore>=0.0.2
 cheetah==2.4.4
 colander==1.0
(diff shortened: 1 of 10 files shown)


marcoceppi:
This is problematic, [Xenial has 2.9.1](http://packages.ubuntu.com/search?keywords=python-requests) so we need to align charm-tools with that.

tvansteenburgh:
Well then theblues will have to unpin, right?

marcoceppi:
Yes, I'll report an issue against that library to follow up that conversation. python-requests has been one of the more reliable libraries as far as backwards compat is concerned so I'm not too worried about the dep pinning. I'll open a separate issue that going forward we should be dep pinning for Xenial and keeping that a target for dependencies so that we can continue to backport charm-tools to xenial over the next two years.

marcoceppi:
LGTM :boat:

requirements.txt
```diff
@@ -1,7 +1,7 @@
 PyYAML==3.11
 blessings==1.6
 bzr>=2.6.0
-charmworldlib>=0.4.2
+libcharmstore>=0.0.2
```
marcoceppi: libcharmstore is 0.0.3

PR #153, opened 2016-03-23
Add --no-local-layers option to charm-build. Fixes #149.

### LINKED ISSUE #149: support a "--no-local-layers" option
On more that one occasion, I have modified a local interface and/or runtime layer, run 'charm build' on something else, and had my local layer in the charm artifact.  It's true I don't read the 'charm build' output as well as I should (at least when it's all green).

I would like an option for 'charm build' that *only* uses interfaces.juju.solutions to pick up included layers.

tvansteenburgh:
Fixes #149


marcoceppi:
I'm sure I'll regret it, but I'll let it slide.

PR #157, opened 2016-03-25
Better feedback to user when command fails

### LINKED ISSUE #126: Improve error reporting when pip fails during wheelhouse creation
http://pastebin.ubuntu.com/15348471/

The output from pip is hidden unless you run with `-l DEBUG` and the stack trace is not a useful error message.  We should catch that and report that it was pip that failed, with the output to explain why.

tvansteenburgh:
Instead of raising a CalledProcessError, write the output of the failed
command, and the command itself, to stderr, then exit with the exit
code of the failed command.

Fixes #126.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ man
 .ropeproject/
 .tox/
 *.pyc
+*.sw[nop]
+.venv/
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -550,7 +550,7 @@ class InstallerTactic(Tactic):
                            "install",
                            "--user",
                            "--ignore-installed",
-                           spec), env=localenv).throw_on_error()()
+                           spec), env=localenv).exit_on_error()()
             self._tracked = []
             # We now manage two classes of explicit mappings
             # When python packages are installed into a prefix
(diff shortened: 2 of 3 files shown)


johnsca:
I'm mostly +1 on this, but it does change the semantics of `throw_on_error` and makes the name misleading (some bit of code might want to catch the exception).  I wonder if it wouldn't be better to catch the exception at a higher level (e.g., `charmtools.build.main`) and convert it to reasonable output there?

Alternatively, add an `exit_on_error` alternative that makes it explicit that it's not just throwing an exception.

tvansteenburgh:
Yeah, I thought of that, but "throw" is never actually what you want in the cases where it's used here. You actually want "tell the user what went wrong and exit". Maybe we should just rename the method?

marcoceppi:
Do we have a consensus on this?

tvansteenburgh:
@marcoceppi @johnsca I'm going to `s/throw_on_error/exit_on_error/g` unless someone objects. There's no reason to keep `throw_on_error` around imo.

marcoceppi:
:+1: would like a +1 from @johnsca but otherwise :boat:

johnsca:
+1 on the rename.  :boat:

PR #162, opened 2016-03-29
Don't build multi-series charms into a series dir

### LINKED ISSUE #115: series in metadata deprecates need to put charms in /series/ directory
We should just build them into `$JUJU_REPOSITORY/<charm>` now, or somewhere more generic

tvansteenburgh:
- If no series given on cmd line and charm is multi-series
  ('series' key in metadata.yaml), don't build in a series dir.
- If no series given on cmd line and charm is not multi-series,
  build in default series dir.
- Added new log.info to show the final destination dir of the
  built charm.
- 'charm_name' property and 'warnings' import removed b/c
  not used anywhere.

Fixes #115.

### DIFF
--- a/charmtools/build/__init__.py
+++ b/charmtools/build/__init__.py
@@ -5,7 +5,6 @@ import json
 import logging
 import os
 import sys
-import warnings
 
 import blessings
 from collections import OrderedDict
@@ -124,6 +123,7 @@ class Builder(object):
     """
     PHASES = ['lint', 'read', 'call', 'sign', 'build']
     HOOK_TEMPLATE_FILE = path('hooks/hook.template')
+    DEFAULT_SERIES = 'trusty'
 
     def __init__(self):
         self.config = BuildConfig()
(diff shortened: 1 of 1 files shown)


charmtools/build/__init__.py
```diff
@@ -188,7 +203,7 @@ def status(self):
     def create_repo(self):
         # Generated output will go into this directory
         base = path(self.output_dir)
-        self.repo = (base / self.series)
+        self.repo = (base / self.series if self.series else base)
```
marcoceppi: What does base / base do in this case?

tvansteenburgh: No, it's "repo = (base/series) or (base)".

marcoceppi: Shouldn't it be base / "builds" per https://github.com/juju/charm-tools/issues/115#issuecomment-201084439

tvansteenburgh: Ugh, I didn't realize that's what you meant by that comment.

marcoceppi: :heart:

marcoceppi:
:boat:

Issue #161, opened 2016-03-29
juju-* is a reserved namespace for actions


marcoceppi:
We need to prevent all actions from being named juju-* similar to interfaces and relations.

PR #164, opened 2016-03-30
lint actions.yaml

### LINKED ISSUE #91: charm proof does not properly lint 'actions.yaml'
When defining actions, there are a few key things to keep in mind and our current tooling does not check the action signatures for compliance of this format.

#### from the tooling developer documentation:

> Note: action names must start and end with lowercase alphanumeric characters, and only contain lowercase alphanumeric characters, the hyphen "-" or full stop"." characters.

cited: https://jujucharms.com/docs/devel/developer-actions

marcoceppi:
This is a first pass at linting actions. We need to do a full lint of the jsonschema to call this complete. However, this will verify the YAML parses, that there are no juju-* actions, and that the action file exists and is executable.

Fixes #91 #161

### DIFF
--- a/charmtools/charms.py
+++ b/charmtools/charms.py
@@ -257,7 +257,9 @@ class Charm(object):
             return lint.lint, lint.exit_code
 
         hooks_path = os.path.join(charm_path, 'hooks')
+        actions_path = os.path.join(charm_path, 'actions')
         yaml_path = os.path.join(charm_path, 'metadata.yaml')
+        actions_yaml_path = os.path.join(charm_path, 'actions.yaml')
         try:
             yamlfile = open(yaml_path, 'r')
             try:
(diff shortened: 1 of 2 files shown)


charmtools/charms.py
```diff
@@ -629,6 +640,39 @@ def validate_payloads(charm, linter):
...
+    :param linter: :class:`CharmLinter` object to which info/warning/error
+        messages will be written
+
+    """
+
+    if not actions:
+        return
+
+    if 'actions' not in actions:
+        linter.err("actions: actions.yaml needs to declare an 'action:' key")
```
tvansteenburgh: declare an 'actions:' key (s is missing from actions)

tvansteenburgh:
Other than that one typo, LGTM.

marcoceppi:
thanks, updated!

PR #166, opened 2016-03-30
clean up descriptions, remove old files

marcoceppi:


marcoceppi:
the fact that tests pass after deleting five files and none of those tests files is a bit sad.

tvansteenburgh:
Did you run all the tests (`make check`) or just unit (`make test`)?

PR #175, opened 2016-04-04
Validate layers for options (fixes #173)

### LINKED ISSUE #173: charm-build should error if defining options for a non-existant layer
When defining layer options according to JSON Schema, such as the following [link describes](http://spacetelescope.github.io/understanding-json-schema/reference/generic.html): i would expect any inheriting layer, that defines an option that doesn't exist in the list to error during  `charm build`

Given in the base layer:

```yaml
defines:
  storage-driver:
    enum: [overlay, aufs, btrfs, zfs]
    default: aufs
    description: |
      Used in conjunction with the storage volume in metadata.yaml.
      Supported providers, when prompted, will allow you to mount a block
      device, formatted for use with the storage driver configured here.
```

And in the inheriting layer:

```yaml
options:
  storage-driver:
    value: schenanigans
```

I expected during the build time, the base schema to be validated. ENUM type should have limited the values to those defined as valid: 'aufs', 'btrfs', 'zfs', or 'overlay' - anything outside of this pre-defined list when provided as a value should have actively prevented the build, or failed it, as it doesn't validate against the defined schema.

johnsca:
Ensure that options are only set for layers that are included.
This also ensures that one doesn't forget to nest the options under the
layer name.

Note: The `additionalProperties` is actually sufficient, but produces a
somewhat obscure error message.

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -374,7 +374,8 @@ class LayerYAML(YAMLTactic):
         super(LayerYAML, self).__init__(*args, **kwargs)
         self.schema = {
             'type': 'object',
-            'properties': {}
+            'properties': {},
+            'additionalProperties': False,
         }
 
     @property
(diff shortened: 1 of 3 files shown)


marcoceppi:
LGTM :boat:

Issue #171, opened 2016-04-04
charm-build should call charm-proof


johnsca:
I don't see any reason for `charm-build` to not immediately call `charm-proof` on the newly built charm.

chuckbutler:
This has been added, yet it appears to be slightly incorrect. See #222

stub42:
For automation, I'd like the return code to be from the build step, not the lint step. Or an option to turn off the automatic lint.

PR #182, opened 2016-04-15
Run `charm proof` immediately after `charm build`

AdamIsrael:
This runs a proof immediately after a successful build, addressing issue #171

### DIFF
--- a/charmtools/build/__init__.py
+++ b/charmtools/build/__init__.py
@@ -23,6 +23,7 @@ from charmtools.build.fetchers import (
 )
 from charmtools import utils
 from .. import repofinder
+from .. import proof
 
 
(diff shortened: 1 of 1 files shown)


charmtools/build/__init__.py
```diff
@@ -673,6 +674,10 @@ def main(args=None):
             build.check_series()
 
         build()
+
+        lint, exit_code = proof.proof(os.getcwd(), False, False)
+        if lint:
+            print("\n".join(lint))
```
johnsca: I think it would be nicer to use `log.warn` to match the colorization of the rest of the output.

johnsca:
+1 :boat:

Issue #174, opened 2016-04-04
proof doesn't catch bundles.yaml -> bundle.yaml


castrojo:
New Juju expects `bundle.yaml`, not `bundles.yaml`. My bundle passed proof but missed this and then the store rejected my upload because it expects `bundle.yaml`.

PR #185, opened 2016-04-18
Catch bundles.yaml -> bundle.yaml

AdamIsrael:
Add a check for bundle.yaml, if juju version is 2.0 or higher

### DIFF
--- a/charmtools/bundles.py
+++ b/charmtools/bundles.py
@@ -4,6 +4,7 @@ import re
 import yaml
 
 from linter import Linter
+from utils import get_juju_version
 import jujubundlelib.validation
 
 
@@ -39,6 +40,13 @@ class BundleLinter(Linter):
                 'https://jujucharms.com/docs/stable/charms-bundles '
                 'for the supported format.')
             return
+
+        # Pushing a bundle to the charm store (Juju 2+) requires a bundle.yaml.
+        ver = get_juju_version()
+        if ver and ver >= 2.0:
+            if not os.path.isfile(os.path.join(
+                    bundle.bundle_path, 'bundle.yaml')):
+                self.err('Juju %s requires a bundle.yaml' % ver)
 
         readmes = glob.glob(os.path.join(bundle.bundle_path, 'README*'))
         if len(readmes) < 1:
--- a/charmtools/bundles.py
+++ b/charmtools/bundles.py
@@ -4,7 +4,6 @@ import re
 import yaml
 
 from linter import Linter
-from utils import get_juju_version
 import jujubundlelib.validation
 
 
--- a/charmtools/bundles.py
+++ b/charmtools/bundles.py
@@ -41,13 +41,6 @@ class BundleLinter(Linter):
                 'for the supported format.')
             return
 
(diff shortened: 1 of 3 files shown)


AdamIsrael:
This addresses issue #174

charmtools/bundles.py
```diff
@@ -40,6 +41,13 @@ def proof(self, bundle):
                 'for the supported format.')
             return
 
+        # Pushing a bundle to the charm store (Juju 2+) requires a bundle.yaml.
+        ver = get_juju_version()
+        if ver and ver >= 2.0:
```
marcoceppi: This isn't tied to a juju version, it should always just be bundle singular.

AdamIsrael: Is bundle.json also deprecated, i.e., bundle.yaml is the only supported
file name going forward?
On Tue, Apr 19, 2016 at 12:52 AM Marco Ceppi <notifications@github.com>
wrote:

> In charmtools/bundles.py
> <https://github.com/juju/charm-tools/pull/185#discussion_r60173042>:
>
> > @@ -40,6 +41,13 @@ def proof(self, bundle):
> >                  'for the supported format.')
> >              return
> >
> > +        # Pushing a bundle to the charm store (Juju 2+) requires a bundle.yaml.
> > +        ver = get_juju_version()
> > +        if ver and ver >= 2.0:
>
> This isn't tied to a juju version, it should always just be bundle
> singular.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/juju/charm-tools/pull/185/files/e28b4512527c33b9ab978471c6add644ad125869#r60173042>
>
-- 
Adam Israel - adam@adamisrael.com
www <http://www.adamisrael.com/> :: Twitter
<http://www.twitter.com/adamisrael> :: Facebook
<https://www.facebook.com/adam.israel> :: Inkpunks
<http://www.inkpunks.com/>

AdamIsrael:
I took the approach of making `bundle.yaml` the only supported filename, and changed the `charm proof` error to note what files it's looking for when checking for a bundle or charm, so the developer has some clue of why proof failed.

charmtools/bundles.py
```diff
@@ -4,6 +4,7 @@
 import yaml
 
 from linter import Linter
+from utils import get_juju_version
```
marcoceppi: we don't need juju_version anymore

AdamIsrael: Whoops! I meant to snip that. Done!

charmtools/proof.py
```diff
@@ -46,7 +46,8 @@ def proof(path, is_bundle, debug):
             try:
                 c = Bundle(path, debug)
             except Exception as e:
-                return ["FATAL: Not a Bundle or a Charm, can not proof"], 200
+                return ["FATAL: No bundle.yaml (Bundle) or metadata.yaml"
+                        "(Charm) found, cannot proof"], 200
```
johnsca: Shouldn't there be a space after `metadata.yaml` or before `(Charm)`?

AdamIsrael: Correct you are!

PR #193, opened 2016-04-27
Enable interface.yaml ignore config option

### LINKED ISSUE #192: interfaces.yaml ignore option doesn't have any effect with charm build
## Checklist

## What version am I running?

`charm tools version: charm-tools==2.1.2`

I am using: `Xenial`

When building a charm, it pulls in interfaces and layers as needed.  There is an optional 'ignore' option that can go in the layer.yaml or interface.yaml to add specs to the DEFAULT_IGNORE list such that it won't copy the ignored files/paths into the built charm.

Unfortunately, it doesn't work with interfaces as the ignore option in interface.yaml is not used.  It does work with layer.yaml files.

This means that we can't ignore (for example) interface specific unit tests that we don't wish to carry over to the final charm.

How to reproduce: create an interface, with some files that you don't wish included.  Add ignore option in interface.yaml.

### What I got

On build the files shouldn't be copied.  The files are copied into the resultant charm.

## Additional info:

Chatting on IRC with cory_fu and bcsaller the possible culprit is at: https://github.com/juju/charm-tools/blob/master/charmtools/build/tactics.py#L159 in that it doesn't use `self.interface.config.ignores` as part of the list of ignores.

ajkavanagh:
This change pulls through an ignore specs in the ignore config file in
the interface.yaml file.

Fixes #192 interfaces.yaml ignore option doesn't have any effect
with charm build

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -156,7 +156,8 @@ class InterfaceCopy(Tactic):
         # directory
         log.debug("Copying Interface %s: %s",
                   self.interface.name, self.target)
-        ignorer = utils.ignore_matcher(self.config.ignores)
+        ignorer = utils.ignore_matcher(self.config.ignores +
+                                       self.interface.config.ignores)
         for entity, _ in utils.walk(self.interface.directory,
                                     lambda x: True,
                                     matcher=ignorer,


marcoceppi:
LGTM :+1:

PR #204, opened 2016-05-03
Add support for `terms` and `resources` keys in metadata.yaml

AdamIsrael:
Add support for new keys introduced in juju 2


marcoceppi:
LGTM :boat:

charmtools/charms.py
```diff
@@ -528,6 +551,48 @@ def schema_type(self, **kw):
...
+    :param charm: dict of charm metadata parsed from metadata.yaml
+    :param linter: :class:`CharmLinter` object to which info/warning/error
+        messages will be written
+
+    """
+    if 'resources' not in charm:
+        return
+
+    if (not isinstance(charm['resources'], dict) or
+            not charm['resources']):
```
johnsca: I feel like a valid but empty dict should not be an error.

AdamIsrael: Not to play the "but everyone else is doing it" card, but that's the same behavior we've adopted for other keys that require a list or dict. If we want to change this, we'd need to update the validation for all the keys of those types.

Issue #186, opened 2016-04-18
The metadata tactic needs de-duplication


mbruzek:
I built a charm from several layers and I see all the maintainers are in the resulting `metadata.yaml` file. Some of this is our fault, we spelled our names or emails differently, but the names that are exactly the same should be de-duplicated.

```yaml
"name": "kubernetes"
"summary": "Kubernetes is an application container orchestration platform."
"description": |
  Kubernetes is an open-source platform for deplying, scaling, and operations
  of appliation containers across a cluster of hosts. Kubernetes is portable
  in that it works with public, private, and hybrid clouds. Extensible through
  a pluggable infrastructure. Self healing in that it will automatically
  restart and place containers on healthy nodes if a node ever goes away.
"tags":
- "containers"
- "layer"
- "containers"
- "networking"
- "infrastructure"
- "security"
- "infrastructure"
"requires":
  "etcd":
    "interface": "etcd"
"provides":
  "dockerhost":
    "interface": "dockerhost"
    "scope": "container"
"peers":
  "certificates":
    "interface": "tls"
"maintainers":
- "Charles Butler <charles.butler@canonical.com>"
- "Matt Bruzek <matthew.bruzek@canonical.com>"
- "Charles Butler <charles.butler@canonical.com>"
- "Matthew Bruzek <matthew.bruzek@ubuntu.com>"
- "Matthew Bruzek <matthew.bruzek@canonical.com>"
- "Charles Butler <charles.butler@canonical.com>"
"subordinate": !!bool "false"
```
The **maintainer** and the **tags** values could be de-duplicated. This could be done by casting them to a set or using a loop to detect if a value has already been written by another method.

marcoceppi:
The problem is, maintainers and tags are lists not sets. This could be easily addressed by converting those two fields in the metadata tactics

johnsca:
I feel like `maintainer` is a case where merging the lists doesn't make sense, and only the charm's `maintainer` should apply.

chuckbutler:
I agree with @johnsca here on this. The top most layer is and should define the actual maintainer.  Just because I had a hand in layer-tls (as an example), doesn't mean I want or have the bandwidth to lend to support the consuming layer(s) that may be using the layer improperly. So attaching my name to the resulting charm is kind of a bad thing to do. 

We'll see layer authors who are putting extra metadata (who knows why, maybe this is for testing with amulet so the layer is standalone) in their layers start to propagate all over the store and own things they had no intention in owning.

[...]

PR #223, opened 2016-06-27
Adding de-duplication for lists in metadata.yaml

mbruzek:
Also adding special case for 'maintainer' and 'maintainers' keys
in metadata.yaml where the only the top layer maintainers are
listed in the built artifact.

## Checklist

 - [ ] Have you followed [Juju Solutions hacking guide?](https://hacking.juju.solutions)

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -187,7 +187,7 @@ class InterfaceCopy(Tactic):
     def lint(self):
         impl = self.interface.directory / self.role + '.py'
         if not impl.exists():
-            log.error('Missing implementation for interface role: %s.py', self.role)
+            log.error('Missing implementation for interface role: %s.py', self.role)  # noqa
             return False
         valid = True
         for entry in self.interface.directory.walkfiles():
@@ -220,7 +220,7 @@ class DynamicHookBind(Tactic):
         for target in self.targets:
             target.parent.makedirs_p()
             target.write_text(template.format(self.name))
-            target.chmod(0755)
+            target.chmod(0o755)
 
     def sign(self):
         """return sign in the form {relpath: (origin layer, SHA256)}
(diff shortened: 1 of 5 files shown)


mbruzek:
According to the discussion on #186 the top layer must be the maintainer (or maintainers). Also found the de-duplication bug for lists in the utils.deepmerge() method.

charmtools/build/tactics.py
```diff
@@ -470,19 +470,29 @@ class MetadataYAML(YAMLTactic):
...
-                 "requires", "provides", "peers"]
+    KEY_ORDER = ["name",
+                 "summary",
+                 "maintainer",
+                 "maintainers",
+                 "description",
+                 "tags",
+                 "requires",
+                 "provides",
+                 "peers"]
```
marcoceppi: nitpick but I'd like to see ] at the same level of `KEY_ORDER` and `"peers"` end in a `,`. This way we can grow the list without having "peers" appear to change just because we added an element

charmtools/build/tactics.py
```diff
@@ -491,6 +501,14 @@ def combine(self, existing):
 
     def apply_edits(self):
         super(MetadataYAML, self).apply_edits()
+        # Remove the merged maintainers from the self.data
+        self.data.pop('maintainer', None)
+        self.data.pop('maintainers', [])
+        # Set the maintainer and maintainers only from this layer.
+        if self.maintainer:
+            self.data['maintainer'] = self.maintainer
+        if self.maintainers:
```
marcoceppi: Since we're already in here wigglin' about, I'd prefer to see us "deprecate" maintainer and instead put everything as a list under maintainers. There by, if there was a maintainer , it'd become a one item list under maintainers.

marcoceppi: I may open a new bug for this, since it's out of scope.

marcoceppi:
LGTM with one nitpick :+1:

charmtools/build/tactics.py
mbruzek: @marcoceppi I considered doing this and discussed it with @johnsca. I came to the conclusion that we should not rename fields, if the author used the "maintainer" key we should keep the same key in the built artifact. Also I could not find anything that said the "maintainer" key was deprecated, to my knowledge it is still a valid key. A new bug would clarify this and I would be happy to change the behavior.

marcoceppi: It's valid, but it's silly to have two IMO. Either way out of scope for this PR

mbruzek:
I addressed the formatting, and ran flake8 which uncovered a few other lint issues that I fixed. Still passes tests.

mbruzek:
I removed the noqa as requested, more invasive but no red flags.

marcoceppi:
:boat:

marcoceppi:
fixes #186

PR #235, opened 2016-07-17
Fix ignore and add exclude support in layer.yaml

### LINKED ISSUE #220: Ignores from lower layers can cause higher layer files to be ignored (dropped)
I've found that not all files survive the charm build process.  In this example, the layer I am building loses its README.md file from the "top" layer.

### Repro
```
git clone https://github.com/openstack/charm-tempest
cd charm-tempest
tox -e build   # or charm build
```

### Take note:  README.md is here.
```
rbeisner@rbc:/tmp/charm-tempest$ ll
total 132
drwxrwxr-x 10 rbeisner rbeisner  4096 Jun  8 16:12 ./
drwxrwxrwt 17 root     root     16384 Jun  8 16:12 ../
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:08 actions/
-rw-rw-r--  1 rbeisner rbeisner   439 Jun  8 16:08 actions.yaml
-rw-rw-r--  1 rbeisner rbeisner  1620 Jun  8 16:08 config.yaml
-rw-rw-r--  1 rbeisner rbeisner   756 Jun  8 16:08 copyright
drwxrwxr-x  8 rbeisner rbeisner  4096 Jun  8 16:08 .git/
-rw-rw-r--  1 rbeisner rbeisner    57 Jun  8 16:08 .gitignore
-rw-rw-r--  1 rbeisner rbeisner    82 Jun  8 16:08 .gitreview
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:08 hooks/
-rw-rw-r--  1 rbeisner rbeisner 30425 Jun  8 16:08 icon.svg
-rw-rw-r--  1 rbeisner rbeisner   108 Jun  8 16:12 layer.yaml
drwxrwxr-x  3 rbeisner rbeisner  4096 Jun  8 16:08 lib/
-rw-rw-r--  1 rbeisner rbeisner   484 Jun  8 16:08 metadata.yaml
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:08 reactive/
-rw-rw-r--  1 rbeisner rbeisner   889 Jun  8 16:08 README.md
-rw-rw-r--  1 rbeisner rbeisner   438 Jun  8 16:08 requirements.txt
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:08 templates/
-rw-rw-r--  1 rbeisner rbeisner   331 Jun  8 16:08 .testr.conf
-rw-rw-r--  1 rbeisner rbeisner   288 Jun  8 16:08 test-requirements.txt
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:08 tests/
-rw-rw-r--  1 rbeisner rbeisner   959 Jun  8 16:08 tox.ini
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:08 unit_tests/
rbeisner@rbc:/tmp/charm-tempest$ cat layer.yaml 
includes: ['layer:openstack', 'interface:keystone-admin']
repo: https://github.com/openstack/charm-tempest
```

```
rbeisner@rbc:/tmp/charm-tempest$ charm build -o build
build: Destination charm directory: build/trusty/tempest
build: Processing layer: layer:openstack
build: Processing layer: layer:basic
build: Processing layer: tempest
build: Processing interface: keystone-admin
```

### Take note:  README.md did not survive to the built artifact.

```
rbeisner@rbc:/tmp/charm-tempest/build/trusty/tempest$ grep README .build.manifest 
    "hooks/relations/keystone-admin/README.md": [

rbeisner@rbc:/tmp/charm-tempest$ ll build/trusty/tempest/
total 136
drwxrwxr-x 11 rbeisner rbeisner  4096 Jun  8 16:14 ./
drwxrwxr-x  3 rbeisner rbeisner  4096 Jun  8 16:13 ../
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:14 actions/
-rw-rw-r--  1 rbeisner rbeisner   493 Jun  8 16:14 actions.yaml
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:14 bin/
-rw-rw-r--  1 rbeisner rbeisner 11951 Jun  8 16:14 .build.manifest
-rw-rw-r--  1 rbeisner rbeisner  2162 Jun  8 16:14 config.yaml
-rw-rw-r--  1 rbeisner rbeisner   756 Jun  8 16:08 copyright
-rw-rw-r--  1 rbeisner rbeisner    57 Jun  8 16:08 .gitignore
-rw-rw-r--  1 rbeisner rbeisner    82 Jun  8 16:08 .gitreview
drwxrwxr-x  3 rbeisner rbeisner  4096 Jun  8 16:14 hooks/
-rw-rw-r--  1 rbeisner rbeisner 30425 Jun  8 16:08 icon.svg
-rw-rw-r--  1 rbeisner rbeisner   322 Jun  8 16:14 layer.yaml
drwxrwxr-x  4 rbeisner rbeisner  4096 Jun  8 16:14 lib/
-rw-rw-r--  1 rbeisner rbeisner   518 Jun  8 16:13 Makefile
-rw-rw-r--  1 rbeisner rbeisner   525 Jun  8 16:14 metadata.yaml
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:14 reactive/
-rw-rw-r--  1 rbeisner rbeisner   438 Jun  8 16:08 requirements.txt
drwxrwxr-x  3 rbeisner rbeisner  4096 Jun  8 16:14 templates/
-rw-rw-r--  1 rbeisner rbeisner   331 Jun  8 16:08 .testr.conf
-rw-rw-r--  1 rbeisner rbeisner   288 Jun  8 16:08 test-requirements.txt
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:14 tests/
-rw-rw-r--  1 rbeisner rbeisner   959 Jun  8 16:08 tox.ini
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:14 unit_tests/
drwxrwxr-x  2 rbeisner rbeisner  4096 Jun  8 16:14 wheelhouse/
```

### Versions (when using charm build)
```
rbeisner@rbc:/tmp/charm-tempest$ charm version
charm 2.1.1-0ubuntu1
charm-tools 2.1.2
```

```
rbeisner@rbc:/tmp/charm-tempest$ apt-cache policy charm
charm:
  Installed: 2.1.1-0ubuntu1
  Candidate: 2.1.1-0ubuntu1
  Version table:
 *** 2.1.1-0ubuntu1 500
        500 http://us.archive.ubuntu.com/ubuntu xenial/universe amd64 Packages
        100 /var/lib/dpkg/status
rbeisner@rbc:/tmp/charm-tempest$ apt-cache policy charm-tools
charm-tools:
  Installed: 2.1.2-0ubuntu4
  Candidate: 2.1.2-0ubuntu4
  Version table:
 *** 2.1.2-0ubuntu4 500
        500 http://us.archive.ubuntu.com/ubuntu xenial/universe amd64 Packages
        500 http://us.archive.ubuntu.com/ubuntu xenial/universe i386 Packages
        100 /var/lib/dpkg/status
```

### Versions (when using tox + venv)
FWIW, I see the same when using tox + venv + pip:
```
rbeisner@rbc:/tmp/charm-tempest$ . .tox/build/bin/activate
(build) rbeisner@rbc:/tmp/charm-tempest$ pip freeze | sort
appdirs==1.4.0
Babel==2.3.4
blessings==1.6
charm-tools==2.1.3
Cheetah==2.4.4
cliff==2.0.0
cmd2==0.6.8
colander==1.3.1
debtcollector==1.4.0
dnspython==1.14.0
ecdsa==0.13
funcsigs==1.0.2
functools32==3.2.3.post2
httplib2==0.9.2
iso8601==0.1.11
Jinja2==2.8
jsonschema==2.5.1
jujubundlelib==0.5.1
keyring==9.0
keystoneauth1==2.8.0
launchpadlib==1.10.3
lazr.authentication==0.1.3
lazr.restfulclient==0.13.1
lazr.uri==1.0.3
libcharmstore==0.0.3
Markdown==2.6.6
MarkupSafe==0.23
monotonic==1.1
msgpack-python==0.4.7
netaddr==0.7.18
netifaces==0.10.4
oauth==1.0.1
os-client-config==1.18.0
oslo.i18n==3.6.0
oslo.serialization==2.7.0
oslo.utils==3.11.0
otherstuf==1.1.0
paramiko==1.17.0
parse==1.6.6
path.py==8.2.1
pathspec==0.3.4
pbr==1.8.1
pkg-resources==0.0.0
positional==1.1.0
prettytable==0.7.2
psutil==1.2.1
pycrypto==2.6.1
pyparsing==2.1.4
python-neutronclient==4.2.0
pytz==2016.4
PyYAML==3.11
requests==2.10.0
requestsexceptions==1.1.3
ruamel.ordereddict==0.4.9
ruamel.yaml==0.11.11
simplejson==3.8.2
six==1.10.0
stevedore==1.14.0
stuf==0.9.16
testresources==2.0.0
theblues==0.3.1
translationstring==1.3
unicodecsv==0.14.1
virtualenv==15.0.2
wadllib==1.3.2
wrapt==1.10.8
wsgi-intercept==1.2.2
zope.interface==4.1.3
```

## Checklist

Top layer root directory files should survive to the built charm.

### What I got

Lost a file.

### LINKED ISSUE #234: charm build AttributeError: 'NoneType' object has no attribute 'combine' on files without extension
## Checklist

## What version am I running?

I ran the following command: `charm version` and got the following output:

```
charm 2.1.1-0ubuntu1
charm-tools 2.1.2
```
Although I have tried to install charm-tools from source to debug this problem.

I am using: Ubuntu 16.04 xenial

I added `00-setup` to the `ignore` field of the layer.yaml and expected the charm build to work.

### What I got

The charm build command failed with a exception and traceback:

```
  File "/usr/local/bin/charm-build", line 9, in <module>
    load_entry_point('charm-tools', 'console_scripts', 'charm-build')()
  File "/home/mbruzek/workspace/git/charm-tools/charmtools/build/builder.py", line 676, in main
    build()
  File "/home/mbruzek/workspace/git/charm-tools/charmtools/build/builder.py", line 519, in __call__
    self.generate()
  File "/home/mbruzek/workspace/git/charm-tools/charmtools/build/builder.py", line 470, in generate
    self.formulate_plan(layers)
  File "/home/mbruzek/workspace/git/charm-tools/charmtools/build/builder.py", line 411, in formulate_plan
    self.plan = self.plan_layers(layers, output_files)
  File "/home/mbruzek/workspace/git/charm-tools/charmtools/build/builder.py", line 326, in plan_layers
    output_files=output_files))
  File "/home/mbruzek/workspace/git/charm-tools/charmtools/build/builder.py", line 322, in <genexpr>
    list(e for e in utils.walk(layer.directory,
  File "/home/mbruzek/workspace/git/charm-tools/charmtools/utils.py", line 349, in walk
    yield (entry, fn(entry, **kwargs))
  File "/home/mbruzek/workspace/git/charm-tools/charmtools/build/builder.py", line 298, in build_tactics
    tactic = current.combine(existing)
AttributeError: 'NoneType' object has no attribute 'combine'
```

Here is the relevant section of the layer.yaml file. It should be noted that none of these file exist in the current layer. I was trying to exclude files coming up from lower layers.

```yaml
ignore:
 - tests/00-setup
 - tests/10-deploy.py
 - tests/10-tls-deploy.py
```
I put charm-tools through a debugger and saw that it was only throwing the exception when processing the `00-setup` file (because it does not have an extension?).

johnsca:
Description of change

## Checklist

Fixes #220 by making ignore apply only to the files provided by layers lower than the one currently being  processed, and by adding an exclude counterpart which applies only to the current layer.

Fixes #234 by refactoring ignore support and making it an error for no tactics to match a file (there should always be a fall-back, such as CopyTactic).


johnsca:
Note: To make the `IgnoreTactic` work, the `trigger` method signature needed to be modified.  I'm fairly sure we don't have any custom tactics in the wild, but if we do, this could potentially break them.

charmtools/build/tactics.py
```diff
@@ -26,12 +26,27 @@ class Tactic(object):
...
-    def __init__(self, entity, current, target, config):
+    @classmethod
+    def get(cls, entity, target, layer, next_config, existing_tactic):
+        """
+        Factory method to get an instance of the correct Tactic to handle the
+        given entity.
+        """
+        for candidate in next_config.tactics + DEFAULT_TACTICS:
+            if candidate.trigger(entity, target, layer, next_config):
+                tactic = candidate(entity, target, layer, next_config)
```
mbruzek: You don't check if tactic is None here. Are you sure this can never be null?

mbruzek:
@johnsca we discussed the two ways charm-tools currently excluded files. I don't see the removal of the existing ignore code. Did I miss it somewhere in here?

charmtools/build/config.py
```diff
@@ -76,39 +73,33 @@ def from_config(cls, config_file, allow_missing=False):
...
+        """
+        if not self.maps:
+            return []
+        return self.maps[0].get('ignore', [])
 
-        for tactic in self.tactics():
-            if tactic.trigger(entity.relpath(bd)):
-                return tactic(target=target, entity=entity,
-                              current=current, config=next_config)
-        return None
```
mbruzek: OK I see the old ignore code removed here.

charmtools/build/builder.py
```diff
@@ -475,8 +476,7 @@ def validate(self):
 
         if not self.manifest.exists():
             return [], [], []
-        ignorer = utils.ignore_matcher(DEFAULT_IGNORES)
```
mbruzek: The old ignore code is removed here.

mbruzek:
I have reviewed this code and it looks sound to me. +1

bcsaller:
Why an ignore tactic at all? In the BulidConfig.tactic method you could return None if the current entity is in the excludes list which won't include a tactic at all. This would be almost identical to the handling of the ignore list there which returns None if next layer would ignore this file. 

Its is very possible I am missing something, but at a glance

```
def tactic(self, entity, current, target, next_config):
        # Produce a tactic for the entity in question
        # These will be accumulate through the layers
        # and executed later
        bd = current.directory
        spec = pathspec.PathSpec.from_lines(pathspec.GitIgnorePattern,
                                            self.excludes)
        p = entity.relpath(bd)
        matches = spec.match_files((p,))
        if p in matches:
            return None

        # Ignore handling
        if next_config:
            spec = pathspec.PathSpec.from_lines(pathspec.GitIgnorePattern,
                                                next_config.ignores)
            p = entity.relpath(bd)
            matches = spec.match_files((p,))
            if p in matches:
                return None

        for tactic in self.tactics():
            if tactic.trigger(entity.relpath(bd)):
                return tactic(target=target, entity=entity,
                              current=current, config=next_config)
        return None


```
seems simpler

Here I include the default ignore list as well, but really that is just saying by default we don't expect our .git/.bzr stuff to make it into the output.

mbruzek:
@bcsaller This actually came up with Issue #234 where I was able to find a case where the current tactic came back as None. https://github.com/juju/charm-tools/blob/master/charmtools/build/builder.py#L298

Using a debugger I was able to find many cases where current and existing were None, but the case where existing was an object and current was None that caused the traceback in the issue.

I added a check for current being None to the if statement and talked with @johnsca about this, he felt that having a None tactic was having no way to handle the file and felt that was not the best way to handle the ignore files.

bcsaller:
I might be misunderstanding something. I mean that if tactic() returns None
it is ignored in the planning phase, see Builder.plan_layers where None
tactics are ignored. This means that the current way to ignore something is
to produce no tactic for it rather than add a new type which does nothing.
*shrug*

charmtools/build/tactics.py
johnsca: candidate is a class not a function, so it is guaranteed to return an instance (this is basically like `IgnoreTactic(entity, target, layer, next_config)`).

johnsca:
@bcsaller Your suggestion is pretty much identical to the existing implementation which suffers from both bugs.

We don't want `tactic()` to return `None` because that ignores the file from the current layer in addition to all below.  Instead, we need to return the correct tactic for the file, but disregard any previous tactic.  There is no way to say "ignore the file from the layers below me, but use the ones I provide" with your implementation (this matters for files that would be merged, or for directories where you want to say, e.g., "ignore all of the tests provided by lower layers, but include the tests I provide").  Further, returning `None` from `tactic()` is indistinguishable from having no tactics match (i.e., we don't know how to handle the file) which is something that we should catch as an error.  Also, a `None` result *wasn't* handled in `plan_layers` for all cases because it was blowing up in `build_tactics`, hence #234.

Your suggestion also doesn't address the issue that ignores (and also now excludes) would be propagated up through any layer that doesn't specify a value, though that only requires the new implementations of the [ignore and excludes properties](https://github.com/juju/charm-tools/pull/235/files#diff-8c32179815b0cbe399e7e2d27b8545daR89).

The main reason I went with `IgnoreTactic` and `ExcludeTactic` was that it seemed like a better design.  It encapsulates the logic for those rules and makes them tactics like anything else, instead of being special-case logic.

This is by no means the smallest nor least invasive change that could have addressed these issues, but it seemed like a reasonable refactor.  However, it does change the signature of the `Tactic` methods, as I mentioned, so if that doesn't seem worth this refactor, if it is disagreed that this refactor is an improvement, or if @marcoceppi is uncomfortable with it for a bug-fix release, I can come up with an alternative with a more minimal diff.

johnsca:
btw, the `DEFAULT_IGNORES` were supposed to be included in both ignores and excludes, and were left out by mistake.  I will update the commit to restore them.

johnsca:
Commit updated to include `DEFAULT_IGNORES`.

bcsaller:
This makes more sense now, thanks for the explanation.

johnsca:
Rebased to master.  Would like to get this merged and included with the next release.

marcoceppi:
Not sure how this one missed the :boat: but it's on it now

PR #244, opened 2016-08-17
--recursive isn't needed for these templates

marcoceppi:
There's no need to do --recursive on these templates since there are no submodules present.

## Checklist


johnsca:
Fair enough, but does it hurt anything to have it in there?

mbruzek:
From what I understand the charm snap does not work when `--recursive` is used on the git command. Marco was going to look into why

johnsca:
Ah, that makes sense.  I'm +1 to merging this.

PR #262, opened 2016-09-24
Indicate when local layers were used and from where

### LINKED ISSUE #258: charm-build -l DEBUG should tell me where it found the layer its using
## Checklist

 - [ ] Provide versions of tools used
 - [ ] Described the feature or ways to replicate the issue

## What version am I running?

I ran the following command: `charm version` and got the following output:

```
charm error
charm-tools 2.1.2

```

I am using:  Ubuntu 16.04 (snap install charm)

## Issue/Feature

When I charm build -l DEBUG i would expect to see some form of output indicating where the layer came from. Either full path on disk that it located or perhaps the github repository for remote layers coming from the interface API server.

This however doesn't appear to be the case, and debugging incorrect layer inclusions is difficult.

johnsca:
Description of change

## Checklist

Fixes #258

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -21,8 +21,9 @@ commands = coverage erase
            coverage run --include=charmtools/* {envbindir}/nosetests {posargs}
            coverage report
 deps =
-    {toxinidir}
+    -e{toxinidir}
     -r{toxinidir}/requirements.txt
+    ipdb
 
 [testenv:lint]
 commands = {envbindir}/flake8 charmtools setup.py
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -21,8 +21,9 @@ commands = coverage erase
            coverage run --include=charmtools/* {envbindir}/nosetests {posargs}
            coverage report
 deps =
-    {toxinidir}
+    -e{toxinidir}
     -r{toxinidir}/requirements.txt
+    ipdb
```
johnsca: I didn't mean to commit these, but they make debugging tests nicer, so I'm leaving it in as a drive-by.

chuckbutler:
Despite the missing code coverage, I am +1 to this change as it clearly indicates where an assembled layer is coming in from. 👍

mbruzek:
+1 from me. We had build time issues that this would have fixed.

PR #263, opened 2016-10-07
use download not install to prevent building locally when building ch…

lutostag:
Do not install things when locally running... Found out that some deps were needed to install locally, using download only gets around that.
## Checklist
- [O] Have you followed [Juju Solutions hacking guide?](https://hacking.juju.solutions) -- umm, that address isn't resolvable for me
- [X] Are all your commits [logically] grouped and squashed appropriately?
- [X] Does this patch have code coverage?
- [O] Does your code pass `make test`? -- cant get the tests to pass locally, hopfully travis will run it for me here ;)

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -753,7 +753,7 @@ class WheelhouseTactic(ExactMatch, Tactic):
     def _add(self, pip, wheelhouse, *reqs):
         with utils.tempdir(chdir=False) as temp_dir:
             # put in a temp dir first to ensure we track all of the files
-            utils.Process((pip, 'install',
+            utils.Process((pip, 'download',
                            '--no-binary', ':all:',
                            '-d', temp_dir) +
                           reqs).exit_on_error()()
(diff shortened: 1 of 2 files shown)


lutostag:
Apparently the version of pip3 on trusty does not have the download directive, so need to fix that first before merging.

chuckbutler:
@lutostag I'm not really sure what this is doing or why it's necessary. Can you outline this fix a bit more for me to understand whats happening here?

lutostag:
@chuckbutler 
I added you to my private charm...
```
$ bzr branch lp:~lutostag/oil-ci/charm-weebl+weasyprint-dep weebl
$ cd weebl; charm build
```
This pulls in some binary python packages -- which need to compile on the host doing the building. This instead pulls down the packages so they can be built on the units themselves (so we can be arch independent). Also means we don't need to have *-dev packages locally for the charm.

You can see that the charm will not be built unless you install libffi-dev locally first.

I was misusing the wheelhouse for non-charm deps and have since gone with a different approach, but if someone wants to use some binary (usually c-compiled binaries) in their pip packages they will hit this.

marcoceppi:
These changes have since been addressed in other pull requests.

PR #266, opened 2016-10-17
Drop unused linting code in the InterfaceCopy tactic

freeekanayaka:
This patch drops some effectively unused logic in the InterfaceCopy.lint method.

AFAIU the goal of the logic being dropped was to calculate a diff between a source and target Python files, however the code seems to have been broken since the very first time it was added (in commit 7a10d7e7db2cd52cd166c92a47faefe4842f1560, the first version of the composer). The reason is that the "entry" and "target" paths that are passed to utils.delta_python_dump will always point to the very same file, so the diff will always be null.

The bug is in the use of the relpath() API:

``` python
>>> import path
>>> entry = path.path("/some/entry")
>>> target_directory = path.path("/the/target")
>>> relpath = entry.relpath(target_directory)
>>> target = target_directory / relpath
>>> entry, target
(path(u'/some/entry'), path(u'/the/target/../../some/entry'))
```

As you can see, "entry" and "target" will point to the same file, as noted above.

Since nobody ever noticed this bug, I think it's safe to assume that the intended behavior is not really very important for the overall functionality of the tool, and since the diffing operation is rather expensive and slows down the program run time, I propose to just drop it.

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -272,19 +272,7 @@ class InterfaceCopy(Tactic):
             log.error('Missing implementation for interface role: %s.py',
                       self.role)
             return False
-        valid = True
-        for entry in self.interface.directory.walkfiles():
-            if entry.splitext()[1] != ".py":
-                continue
-            relpath = entry.relpath(self._target.directory)
-            target = self._target.directory / relpath
-            if not target.exists():
-                continue
-            unchanged = utils.delta_python_dump(entry, target,
-                                                from_name=relpath)
-            if not unchanged:
-                valid = False
-        return valid
+        return True
 
 
 class DynamicHookBind(Tactic):


johnsca:
@bcsaller I think the intent of that check was perhaps to disallow overriding of interface layers?  I'm not clear on it, but I can see why we might want to enforce that to ensure compatibility as is the purpose of interface layers.

bcsaller:
What we are really trying to do here is find a way to tell developers,
something you depend on has changed, when its an interface we want to
indicate, there are new states or features you could be using. Even if this
code path worked it doesn't properly implement that intent. While part of
me thinks that just fixing what we have today is better than nothing if my
use-case is still felt to be an important one we should have a session
about it at the next sprint.

freeekanayaka:
Gut feeling: it seems that the intent of signalling that a dependency
has changed upstream is a general problem of any vendored
dependency. By that reasoning we should consider applying
similar mechanism for any dependency that you declare, say,
in your wheehouse.txt, no?

It seems that layers are somehow a "parallel" vendoring mechanism
alongside the regular Python (e.g. pip). I'm then wondering if layers
shouldn't be treated like any other Python dependency, that you declare
in wheelhouse.txt. Overriding a certain layer can be then treated more
explicitly as regular and explicit forking (though that's something I would
consider as temporary band aid for fixing bugs or adding behavior that
should then be pushed upstream, eliminating the fork).

Please let me know if there's some fundamental misunderstanding of
the idea on my part.

bcsaller:
This thinking has merit but to my mind there is a difference in a
traditional dep, like say charm-helpers and what layers provide. charm
build is a build time operation that takes care of things like boilerplate,
and wiring certain types of code (interfaces for example) into a runnable
whole. It is a more specialized tool than a package manager because it
attempts to be both content and process aware. By content I mean (again)
things like infterfaces where we can introduce special handling at build
time. By process aware we hoped that the whole charm tools suite could help
shepherd a new charm author to a successful working charm quickly, but also
help them keep that charm up to date within the larger ecosystem.

I admit there is a great deal of room for improvement here, perhaps some of
this is better done through a more sophisticated interfaces.juju.solutions
site which could track deps so as indicate impacted charms and layer at the
point at which a change to the index is being made. In truth though it is
more complex that even that, we often point to a repo which will change
through an index which wont so now it's a slightly more complex problem.
This process of keeping a charm up to date though has a content aware
signaling. The idea being not just that an interface has rev'd a version
but if there are any new states or data to respond to.

It feels like this has a home in charm-tools and not as much through
something like pip. Either way I hope people feel it is worth a session at
the sprint. Thanks for tracking this down and bringing up the issue.

PR #271, opened 2016-10-30
Snap version update

marcoceppi:
This makes it so everyone using charm-version from within a snap will actually get a result.

## Checklist

 - [ X ] Are all your commits [logically] grouped and squashed appropriately?
 - [ X ] Does your code pass `make test`?

### DIFF
--- a/charmtools/version.py
+++ b/charmtools/version.py
@@ -1,4 +1,5 @@
 
+import os
 import pkg_resources
 import argparse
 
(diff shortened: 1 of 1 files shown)


marcoceppi:
```
$ which charm
/snap/bin/charm
```
And the output
```
$ charm version
charm 2.2.0
charm-tools 2.1.2
```

PR #284, opened 2016-11-15
Create directory, regardless of entity

### LINKED ISSUE #283: Charm build removes empty directories
## Checklist

 - [x] Confirmed this is an issue with charm-tools, not charmstore-client
 - [x] Provide versions of tools used
 - [x] Described the feature or ways to replicate the issue

## What version am I running?

```
$ charm version
charm 2.2.0-0ubuntu1~ubuntu16.04.1~ppa2
charm-tools 2.1.7
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.1 LTS
Release:	16.04
Codename:	xenial
```
## Issue/Feature

I have empty directory in my layer (on purpose):
`files/archive-mirror-scripts/filelists/`

I composed charm using:
`charm build -s xenial`

But above unhelpfully removed the directory without any warning message. In composed charm directory:
`ls: cannot access 'files/archive-mirror-scripts/filelists': No such file or directory`

I expected `charm build` to preserve empty directories, I created them on purpose, it's much easier to rsync directory tree than create it from python code.
If there is a good reason to modify layers on the fly `charm build` should make it very clear to avoid surprises.

### What I got

`charm build` silently modified my layer without printing any error, warning or info messages. This resulted in broken charm

marcoceppi:
fixes #283

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -198,7 +198,8 @@ class ExcludeTactic(Tactic):
 class CopyTactic(Tactic):
     def __call__(self):
         if self.entity.isdir():
-            return
+            log.debug('Creating %s', self.target_file)
+            return self.target_file.makedirs_p()
         target = self.target_file
         log.debug("Copying %s: %s", self.layer_name, target)
         # Ensure the path exists


johnsca:
:+1:

johnsca:
Good catch.  :+1: again

PR #298, opened 2017-01-05
Remove duplicates from series field in metadata tactic fixes #297

### LINKED ISSUE #297: Charm build creates duplicated `series` entries
## Checklist

## What version am I running?

```
charm 2.2.0-0ubuntu1~ubuntu16.04.1~ppa2
charm-tools 2.1.9

```

I am using: Ubuntu 16.04

When building a charm that has two layers that both include a `metadata.yaml` file with the `series` field, the `series` of the top-level layer should be used in the final Charm

### What I got

The `series` field got merged, resulting in duplicated series:

    series: ['xenial', 'trusty', 'xenial', 'trusty']

Charms with series fields such as this break the charm store. See https://lists.ubuntu.com/archives/juju/2017-January/008416.html

marcoceppi:
This de-duplicates metadata series and fixes #297

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -593,6 +593,7 @@ class MetadataYAML(YAMLTactic):
         "maintainers",
         "description",
         "tags",
+        "series",
         "requires",
         "provides",
         "peers",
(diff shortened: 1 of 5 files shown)


johnsca:
Do you think you could add a `trusty` series to the mysql layer's `metadata.yaml` to ensure that de-dupe works across layers as well (since you ran in to that while testing)?

Issue #241, opened 2016-08-03
Build charms making invalid legal claims


stub42:
## Checklist

 - [X] Confirmed this is an issue with charm-tools, not charmstore-client
 - [X] Provide versions of tools used
 - [X] Described the feature or ways to replicate the issue

## What version am I running?

I ran the following command: `charm version` and got the following ouput:

```
charm 2.1.1-0ubuntu1
charm-tools 2.1.2
```

I am using: Ubunut

## Issue/Feature

Build needs to handle copyright files correctly.

- If a charm does not include a copyright file, when composed one from one of its layers is used instead. charm proof doesn't complain, and invalid legal assertions are made

- When composing a charm, copyright statements from layers are thrown away, possibly in violation of licenses like BSD

Instead, charm build needs to collate and keep the copyright files in the composed charm. Possibly by just renaming them copyright.$layer, or sticking them in a licenses directory.

### I expect/expected the following

No bugs and world peace.

### What I got

A face palm from legal.

chuckbutler:
Yeah, we had some preliminary discussions about this early on, and I'm pretty sure we punted on it at the time. There was talk of license linting and an even more complex implementation that scanned for complimentary license options... I'm not sure what happened to that thread.

stub42:
I'd start simple and iterate, since perfect is impossible in this case.

1. Keep copyright statements from layers
1. Move to machine readable format
1. Report licenses at build time, so the charmer can make an informed decision
1. Magic system to detect and block conflicts at build time based on the charmers current legal jurisdiction.
1. A pony.

This issue is calling out for the first step.

PR #318, opened 2017-04-26
Add tactic for copyright

galgalesh:
This PR is the start for a solution to https://github.com/juju/charm-tools/issues/241

I created a tactic that renames each copyright file to `copyright.<type>-<name>` during `charm build`. The copyright file of the top level layer remains untouched.

The `HACKING.TXT` file seems a bit out of date so I updated it.

## Checklist

 - [x] Have you followed [Juju Solutions hacking guide?](https://hacking.juju.solutions) **PS: that URL doesn't resolve**


chuckbutler:
Assigning to Marco and Cory as I know they have experience in these tactic classes. The code looked fine to me but I'm not comfortable casting a +1 without a minimum test suite testing the functionality of the code on some test files.

galgalesh:
@chuckbutler, @johnsca , @marcoceppi 

I added two tests for this tactic; it checks if the copyright files are indeed renamed correctly.

chuckbutler:
Nice :) :+1:   I like how the tests helped cleanup that method block in `_call()`

marcoceppi:
I think this is a fine start, I think we may run into an issue for charms built with more than a few layers where the files start to pile up. This code looks good though and I'm inclined to merge. Going to have a think about it for a bit longer but will circle back before the end of week

galgalesh:
Another option is to use the Debian machine-readable `copyright` format to specify which licenses apply to which files but that would open a can of worms:

- Which licenses apply to which files?
- What to do with merged files?
- What to do with overwritten files?
- What to do with `copyright` files that don't really follow the debian `copyright` specification (almost all of them)
- ...

marcoceppi:
I agree a machine-readable format would be preferred. I don't think it's feasible at this point and I'd like to start moving this PR forward instead and work on iterating towards a more sane format.

> Which licenses apply to which files?

We already know which license applies to which files as it's per layer and each layer is known at compile time.

> What to do with merged files?

Merged files is hard, but I think we should really /really/ push for any lower layers being an open license like Apache or MIT. Which would help resolve a lot of these merge issues

> What to do with overwritten files?

Overwritten files means copyright is assigned to the layer that just overwrote it.

> What to do with copyright files that don't really follow the debian copyright specification

This one is a tough one, and I think we should pick a format and then enforce it over a few months time. Work with lower layer authors to get them converted. Lets move this PR forward and open a follow up issue to lay out a roadmap for copyrights

stub42:
If charms become two unwieldy with many COPYRIGHT files in the top level, we can shuffle them into a subdirectory easily enough. I don't see a need to do that yet.

galgalesh:
@johnsca 

This PR seems to have been lost in the weeds. Can you review it please?

Issue #305, opened 2017-02-24
Charm build output dir is different depending on wher you run it.


galgalesh:
## Checklist

 - [X] Confirmed this is an issue with charm-tools, not charmstore-client
 - [ ] Provide versions of tools used
 - [ ] Described the feature or ways to replicate the issue

## What version am I running?

I ran the following command: `charm version` and got the following ouput:

```
charm 2.2.0-0ubuntu1~ubuntu16.04.1~ppa2
charm-tools 2.2.0
```

I am using: Ubuntu 16.04

## Issue/Feature

`charm build` has a different output directory depending on how you supply the layer to build.

Reproduce: 

```bash
cd $LAYER_PATH
`charm build openvpn`
# destination dir is $JUJU_REPOSITORY/builds/openvpn

cd $LAYER_PATH/../
`charm build openvpn`
# destination dir is $JUJU_REPOSITORY/trusty/openvpn

cd $LAYER_PATH/../
`charm build $LAYER_PATH/openvpn`
# destination dir is $JUJU_REPOSITORY/builds/openvpn
```

### I expect/expected the following

All these commands have the same output dir

### What I got

The output dir is different

galgalesh:
@petevg also has this issue.

johnsca:
We should revisit this and make the output directory consistent.  We should also remove the default behavior when `$JUJU_REPOSITORY` is not set to build inside the current directory, since that will almost always lead to incorrect results.

Issue #259, opened 2016-09-21
charm build outputs to trusty if top-level layer doesn't include series


johnsca:
## Checklist

 - [X] Confirmed this is an issue with charm-tools, not charmstore-client
 - [X] Provide versions of tools used
 - [X] Described the feature or ways to replicate the issue

## What version am I running?

I ran the following command: `charm version` and got the following ouput:

```
charm 2.2.0-0ubuntu1~ubuntu16.04.1~ppa2
charm-tools 2.1.4
```

I am using: Ubuntu Xenial

## Issue/Feature

If a top-level charm layer is depending on a base layer to provide the supported series (e.g., if it's a thin wrapper over the base charm) and thus has no `series` field in its own `metadata.yaml`, when you do `charm build` it doesn't look in the built `metadata.yaml` for a series, so your multi-series charm ends up in `/trusty/`.

When Juju 2.0 is officially released, `/builds/` should be the default, so that the output dir is always consistent.

Issue #213, opened 2016-05-30
output directory is a lie


stub42:
Otherwise, if you have a feature request or issue with any other charm command please delete the above and continue below

## Checklist

 - [X] Confirmed this is an issue with charm-tools, not charmstore-client
 - [X] Provide versions of tools used
 - [X] Described the feature or ways to replicate the issue

## What version am I running?

I ran the following command: `charm version` and got the following ouput:

```
charm 2.1.1-0ubuntu1
charm-tools 2.1.2
```

I am using: Ubuntu 16.04

## Issue/Feature

'charm build' is generating charms into the deprecated Juju 1.x repository format. Tested using an empty layer created by 'charm create -t reactive-python'

### I expect/expected the following

'charm build -o somedirectory' should generate the charm into 'somedirectory' as documented.

### What I got

'charm build -o somedirectory' generates the charm into somedirectory/trusty/charmname.

stub42:
For even more confusion, if the charm declares its supported series 'charm build -o somedirectory' generates the charm into somedirectory/builds/charmname. Unless the -s argument is also passed, in which case it goes into somedirectory/series/charmname.

The only work around seems to be to construct a nest of symlinks to trick 'charm build' into generating the charm into the desired directory.

PR #320, opened 2017-05-03
Normalize output directory

galgalesh:
The `charm build` command's output directory seems to depend on the position of the stars. **This PR normalizes the output directory with a few simple rules and explains the behavior in the help text.**

 1. `output-dir` defaults to `$JUJU_REPOSITORY`.
 2. *If `$JUJU_REPOSITORY` is unset, `output-dir` defaults to the directory of the top-level layer.*
 3.  Build to `<output-dir>/builds/<name>` by default.
 4.  Build to `<output-dir>/<series>/<name>` when series is explicitly defined as argument.
 5. *Do the in-place-upgrade magic when the stars are right*

This fixes a couple of issues:

- Charm build output dir is different depending on where you run it: https://github.com/juju/charm-tools/issues/305
- Output directory is a lie: https://github.com/juju/charm-tools/issues/213
- Charm build outputs to trusty if top-level layer doesn't include series: https://github.com/juju/charm-tools/issues/259

PS: *I'd also like to remove 2. and 5. but this might give some issue with backwards compatibility. In my opinion, if $JUJU_REPOSITORY isn't set, and no output-dir is specified, charm build should display an error message explaining this. This will greatly help new users who haven't set $JUJU_REPOSITORY. @johnsca , @marcoceppi : Thoughts?*

## Checklist

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -160,15 +160,13 @@ class Builder(object):
         in <output_dir>/builds/<charmname> when no series is specified or
         <output_dir/<series>/<charmname> when series is specified."""
         if not self._output_dir:
-            self._output_dir = path(os.environ.get('JUJU_REPOSITORY'))
+            self._output_dir = path(os.environ.get('JUJU_REPOSITORY', self.charm)).abspath()
         return self._output_dir
 
     @output_dir.setter
     def output_dir(self, value):
         if value:
             self._output_dir = path(value)
-        else:
-            self._output_dir = path(os.environ.get('JUJU_REPOSITORY'))
 
     @property
     def deps(self):
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -131,6 +131,7 @@ class Builder(object):
         self._name = None
         self._charm = None
         self._top_layer = None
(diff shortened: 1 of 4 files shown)


johnsca:
The only reason we included the series in the path was for Juju 1.x which required the series to be part of the path.  I think we're far enough in to 2.x that we can push for dropping it entirely.  I would also go so far as to say we should drop the `builds` portion of the path, since that has nothing to do with how Juju deploys charms and actually makes it harder to deploy the charms once built.  I'm also +1 on requiring the output directory to be explicit, either from the CLI or the env var.

I think it's worth raising this on the mailing list, as I don't think there will be too much push back against making this more consistent and sensible.

chuckbutler:
I agree Cory. Lets drop the dead code paths and push forward with juju 2 standards.

marcoceppi:
In this same vein, we should make the series key a required component of
metadata.yaml

stub42:
:+1: 

This will break several of my Makefiles, but I'd rather fix them and have nicer paths than live with the techdebt.

galgalesh:
Closing this PR since it has diverged too much from master

PR #330, opened 2017-07-19
snapcraft

battlemidget:
Updated snapcraft for charm-tools

@tvansteenburgh @johnsca - This builds and run's but we should check to make sure all the subcommands work and do what you expect (mainly things like git support, terms as those are additional parts in the snap)


marcoceppi:
Merging, and enabling in build.snapcraft

marcoceppi:
https://build.snapcraft.io/user/juju/charm-tools I'll open a PR for badges

PR #333, opened 2017-07-21
Modify snap to use strict confinement

johnsca:
The classic confinement was causing the environment to bleed into the charm in some cases causing problems such as `charm build` failing when used within Travis.  This enables strict confinement.

Caveats:
* Can only build to / from directories under $HOME
* The strictly confined snap can't read existing auth cookies, so when switching you will need to re-login (once)
* This also removes `charm test` / `juju test` since they have been deprecated for some time and don't work with strict confinement


stub42:
You will want to test a charm with dotfiles too - the 'home' interface doesn't allow access to 'hidden' files, making it pretty much useless for this sort of thing.

PR #334, opened 2017-07-21
Add bzr to build_packages to fix build.snapcraft.io

johnsca:


johnsca:
close in favor of #335

PR #338, opened 2017-08-24
Add option for wheelhouse overrides

johnsca:
To facilitate CI and testing of charms with new versions of Python dependencies, particularly those from base layers, you can now provide a wheelhouse.txt file via the `charm-build` CLI that will be applied
after all others which can thus override any dependencies from lower layers.

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -332,7 +332,7 @@ class Builder(object):
         if self.wheelhouse_overrides:
             existing_tactic = output_files.get('wheelhouse.txt')
             output_files['wheelhouse.txt'] = WheelhouseTactic(
-                self.wheelhouse_overrides,
+                str(self.wheelhouse_overrides),
                 self.target,
                 layers["layers"][-1],
                 next_config,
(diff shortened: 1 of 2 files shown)


johnsca:
@ryan-beisner

charmtools/build/builder.py
```diff
@@ -328,6 +329,16 @@ def plan_layers(self, layers, output_files):
...
+        if self.wheelhouse_overrides:
+            existing_tactic = output_files.get('wheelhouse.txt')
+            output_files['wheelhouse.txt'] = WheelhouseTactic(
+                self.wheelhouse_overrides,
+                self.target,
+                layers["layers"][-1],
+                next_config,
+            )
+            if existing_tactic is not None:
+                output_files['wheelhouse.txt'].combine(existing_tactic)
```
marcoceppi: Should we combine or just be an overwrite?

marcoceppi: Nevermind, thinking about it combine makes more sense

johnsca: Combine merges the files.  In fact, we end up with both versions of the files in the wheelhouse, which is also something that can happen with multiple layers referencing the same dep.  It would be great to be able to resolve that up front, but that would require reimplementing much of pip's dependency resolution, so I'm disinclined to even try.  It does generally do the right thing at the end, but it may be something we need to address eventually.

johnsca:
@marcoceppi I can look at adding / extending tests tomorrow.

marcoceppi:
Thanks @johnsca, looks great! Should be in edge shortly.

PR #341, opened 2017-09-07
Migrate layer index to GitHub repo

### LINKED ISSUE #324: Migrate from interfaces.juju.solutions to a GitHub repo
The layer index site works, but was originally intended to be an interim to finding a more permanent solution.   It has not, however, been improved or replaced.  It is also confusing for users in that it is separate from both the charm store and the code base.  Instead of maintaining an active index site & API, we should use a GitHub repo to manage the layer index.  This would allow us to use PRs for managing the index, giving more visibility to the community when layers are added, modified, or removed.

I think we should probably have a YAML file per record in the current database, and we might want to include some script or hook for updating a Markdown index (perhaps a subsection of the README.md) with clickable links to the repos referenced in the YAML files.

johnsca:
This resolves #324 by moving the layer index to https://github.com/juju/layer-index

### DIFF
--- a/charmtools/build/fetchers.py
+++ b/charmtools/build/fetchers.py
@@ -27,12 +27,12 @@ fetchers.FETCHERS.insert(0, RepoFetcher)
 
 
 class InterfaceFetcher(fetchers.LocalFetcher):
-    # XXX: When hosted somewhere, fix this
-    INTERFACE_DOMAIN = "http://interfaces.juju.solutions"
+    INTERFACE_DOMAIN = "https://raw.githubusercontent.com/" \
+            "juju/layer-index/master/"
     NAMESPACE = "interface"
     ENVIRON = "INTERFACE_PATH"
     OPTIONAL_PREFIX = "juju-relation-"
-    ENDPOINT = "/api/v1/interface"
+    ENDPOINT = "interfaces"
     NO_LOCAL_LAYERS = False
 
     @classmethod
@@ -58,8 +58,9 @@ class InterfaceFetcher(fetchers.LocalFetcher):
             if name.startswith(cls.OPTIONAL_PREFIX):
                 choices.append(name[len(cls.OPTIONAL_PREFIX):])
             for choice in choices:
-                uri = "%s%s/%s/" % (
+                uri = "%s%s/%s.json" % (
                     cls.INTERFACE_DOMAIN, cls.ENDPOINT, choice)
+                print(uri)
                 try:
                     result = requests.get(uri)
                 except:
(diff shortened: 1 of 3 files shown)


johnsca:
[(2) interfaces.juju.solutions to GitHub](https://trello.com/c/ztZc4lhl/5-2-interfacesjujusolutions-to-github)

charmtools/build/fetchers.py
```diff
@@ -112,10 +113,11 @@ def fetch(self, dir_):
 
 
 class LayerFetcher(InterfaceFetcher):
-    INTERFACE_DOMAIN = "http://interfaces.juju.solutions"
+    INTERFACE_DOMAIN = "https://raw.githubusercontent.com/" \
```
marcoceppi: Would it be worth moving this from away from raw to a github site? raw access is generally discouraged.

battlemidget:
> One piece of feedback. I think we should move to a github site, and see if we can't make it backwards compatible (as in point interfaces.juju.solutions to it, etc)

Eventually we could do a github site/page and have some clientside javascript to render the list of layers/interfaces to improve discoverability.

johnsca:
The only thing not backwards compatible is the .json suffix, and that could be handled in the redirect pretty easily.  I'm not against the data being served from a GitHub site / page, but I want to make sure that the repo update flow (PRs, issues, etc) are standard.  I'm not sure how to reliably get the data into a page from the main repo.

**Edit:** The rewrite rule could be something like:

```
RewriteRule ^/api/v1/([^/]*)/(.*)  https://raw.githubusercontent.com/juju/layer-index/master/$1s/$2.json
```

johnsca:
Also, can you point to documentation indicating that the raw URL is discouraged?  It's surfaced directly via the GitHub site and I haven't seen anything suggesting that it's discouraged.

battlemidget:
Additionally, there seems to be full blown sites dedicated to serving the raw files with correct headers: https://rawgit.com/

marcoceppi:
gh-pages are just static files in a branch. If you remove the master branch and create a `gh-pages` branch, set that as the primary. Then pull requests, issues, etc all work the same, but the files are published by gh in a way meant to be consumed.

Finally, these workaround sites exist because raw.githubusercontent is designed for people using the web interface to view code, not for serving files. Certainly, we don't fit the description as a "heavy user", but they note that there is a [toll to rendering these files](https://github.com/blog/1482).

johnsca:
@marcoceppi Fair enough.  Consider me convinced.  I'll look into reconfiguring the repo and updating this PR.

johnsca:
On the plus side, we tested the existing `charm-build` with the redirect, and it's seamlessly backwards compatible.

PR #345, opened 2017-09-21
charm-build output fixes

johnsca:
Remove pront from #341, replace with debug log messages
Show source of local interfaces like we do with layers


marcoceppi:
Haha, pront.

Issue #327, opened 2017-06-30
charm create from snap fails: OSError: [Errno 2] No such file or directory: '/tmp/tmpOx6ca0'


dannf:
## Checklist

 - [X] Confirmed this is an issue with charm-tools, not charmstore-client
 - [X] Provide versions of tools used
 - [X] Described the feature or ways to replicate the issue

## What version am I running?

I ran the following command: `charm version` and got the following ouput:

```
$ charm version
charm 2.2.2
charm-tools 2.2.1
```

I am using: Debian GNU/Linux

## Issue/Feature
charm create fails.
### I expect/expected the following
No crash.

### What I got
```
$ charm create java-browser
INFO: Using default charm template (reactive-python). To select a different template, use the -t option.
INFO: Generating charm for java-browser in ./java-browser
INFO: No java-browser in apt cache; creating an empty charm instead.
Cloning into '/tmp/tmpOx6ca0'...
fatal: Unable to find remote helper for 'https'
Traceback (most recent call last):
  File "/snap/charm/15/bin/charm-create", line 11, in <module>
    sys.exit(main())
  File "/snap/charm/15/lib/python2.7/site-packages/charmtools/create.py", line 98, in main
    generator.create_charm()
  File "/snap/charm/15/lib/python2.7/site-packages/charmtools/generators/generator.py", line 77, in create_charm
    self._cleanup(tempdir)
  File "/snap/charm/15/lib/python2.7/site-packages/charmtools/generators/generator.py", line 127, in _cleanup
    shutil.rmtree(tempdir)
  File "/snap/charm/15/usr/lib/python2.7/shutil.py", line 239, in rmtree
    onerror(os.listdir, path, sys.exc_info())
  File "/snap/charm/15/usr/lib/python2.7/shutil.py", line 237, in rmtree
    names = os.listdir(path)
OSError: [Errno 2] No such file or directory: '/tmp/tmpOx6ca0'
```

deephack1982:
Can confirm the same here. Juju and Charm tools installed from snap packages.

Ubuntu 17.04
charm 2.2.2
charm-tools 2.1.2

`$ charm create -v paradox-linux-buildslave-puppet
INFO create.py: Using default charm template (reactive-python). To select a different template, use the -t option.
INFO generator.py: Generating charm for paradox-linux-buildslave-puppet in ./paradox-linux-buildslave-puppet
INFO utils.py: No paradox-linux-buildslave-puppet in apt cache; creating an empty charm instead.
[...]

PR #347, opened 2017-09-26
Validate that input and output paths can be accessed by snap

### LINKED ISSUE #342: charm build should display an error if it can't build the charm
## Checklist

## What version am I running?

```
charm 2.2.2
charm-tools 2.1.2
```

I am using: Ubuntu 17.10, with 'charm' installed as a snap from the stable channel:

```
Name              Version    Rev   Developer  Notes
charm             2.2        17    charms     -
```

## Issue/Feature

After doing some reading, I understand that the latest charm snap is now strictly confined, and that it therefore can't write outside my homedir. However, I do think that if an unsuspecting user requests that it do so, the user should be warned that this isn't possible.

The output certainly shouldn't suggest that the charm has been built and written, when it hasn't at all.

I ran `charm build -o /tmp -n <foo>` expecting to build a charm and write it to `/tmp/builds/<foo>`

### What I got

```
$ charm build -o /tmp -n <foo>
build: Destination charm directory: /tmp/builds/<foo>
build: Processing layer: <layer_1>
build: Processing layer: <layer_1>
build: Processing layer: <foo> (from .)
build: Processing interface: <interface_1>
build: Processing interface: <interface_2>
build: Processing interface: <interface_3>
build: Processing interface: <interface_4>
build: Processing interface: <interface_5>
proof: OK!
$ ls -l /tmp/builds/
ls: cannot access '/tmp/builds': No such file or directory
$ 
```

johnsca:
Fixes #342 and partially addresses #327 by rejecting attempts to build or create into or from directories outside of `$HOME` with a clear error message.

## Checklist

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -547,8 +547,29 @@ class Builder(object):
                 od = repo
         elif ":" in od:
             od = od.basename
-        log.info("Composing into {}".format(od))
         self.output_dir = od
+
+    def _check_path(self, path_to_check):
+        home_dir = os.path.expanduser('~')
+        if home_dir == '~':
+            raise BuildError('Could not determine home directory')
+        return os.path.abspath(path_to_check).startswith(home_dir)
+
+    def check_paths(self):
+        paths_to_check = [
+            self.output_dir,
+            self.wheelhouse_overrides,
+            os.environ.get('JUJU_REPOSITORY'),
+            os.environ.get('LAYER_PATH'),
+            os.environ.get('INTERACE_PATH'),
+        ]
+        for path_to_check in paths_to_check:
+            if path_to_check and not self._check_path(path_to_check):
+                raise BuildError('Due to snap confinement, all paths must be '
+                                 'under your home directory, including the '
+                                 'build output dir, JUJU_REPOSITORY, '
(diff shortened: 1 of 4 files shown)


johnsca:
Not sure why codacy is complaining about the use of `assert` in a test.  That's kind of the point.

johnsca:
[(2) charm build should display an error if it can't build the charm (#342)](https://trello.com/c/k4V6iAWu/75-2-charm-build-should-display-an-error-if-it-cant-build-the-charm-342)

charmtools/build/builder.py
```diff
@@ -547,9 +547,30 @@ def normalize_outputdir(self):
...
+        paths_to_check = [
+            self.output_dir,
+            self.wheelhouse_overrides,
+            os.environ.get('JUJU_REPOSITORY'),
+            os.environ.get('LAYER_PATH'),
+            os.environ.get('INTERACE_PATH'),
+        ]
+        for path_to_check in paths_to_check:
+            if path_to_check and not self._check_path(path_to_check):
+                raise BuildError('Due to snap confinement, all paths must be '
```
marcoceppi: This isn't entirely true, `$SNAP_DATA_DIR` is also valid, though less likely to be used.

PR #358, opened 2017-10-26
Bug/357 versioning

johnsca:
Clean up README and use RST so it looks ok on PyPI.
Make the version tags in git authoritative for the version, and report version for non-stable channels as tagged version plus info on how far diverged it is (number of new commits, and sha of last commit in release).

## Checklist


VERSION
```diff
@@ -0,0 +1 @@
+2.2.0  # do not edit, managed by git tags
```
johnsca: Maybe this should be in `.gitignore`?

johnsca:
Codacy is complaining about the use of subprocess without offering any info on why it's complaining.  We need subprocess to call git, and we're not doing anything with the output aside from populating the version, so it's not a security concern.  I can't find a way to tell Codacy that it's fine, so I guess we just ignore it?

snap/snapcraft.yaml
```diff
@@ -1,5 +1,6 @@
 name: charm
-version: 2.2
+version: git
+version-script: scripts/git-version
```
kwmonroe: These `version` and `version-script` bits are foreign to me, but reading this gives me some confidence that you're doing it right:

https://forum.snapcraft.io/t/snapcraft-version-scripts/106/44

PR #359, opened 2017-10-26
Fixed version handling

johnsca:
SNAPCRAFT_GRADE doesn't do what I thought


charmtools/version.py
```diff
@@ -39,15 +44,20 @@ def charm_version():
     return charm_ver
 
 
-def charm_tools_version():
-    return pkg_resources.get_distribution("charm-tools").version
+def charm_tools_version(format):
```
kwmonroe: codacy hates you for `format`, if that's something you're concerned about.

charmtools/git_version.py
```diff
@@ -0,0 +1,70 @@
...
+from pkg_resources import parse_version
+
+
+git_cmd = ['git', 'describe', '--tags', '--long']
+version_filename = os.path.join(os.path.dirname(__file__), '..', 'VERSION')
+
+
+def get_version_info():
+    try:
+        version, gitn, gitsha = check_output(git_cmd).strip().rsplit('-', 2)
```
battlemidget: Can we just ignore this error from codacy?

PR #378, opened 2017-11-29
Add github.com/juju/plans-client commands.

cmars:
Add plan management commands from github.com/juju/plans-client to the snap.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -16,24 +16,12 @@ apps:
 parts:
   charm:
     source: https://github.com/juju/charmstore-client.git
-    source-tag: 2.2.2
+    source-tag: 2.2.3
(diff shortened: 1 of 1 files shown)


marcoceppi:
LGTM :+1: can you please update this with the proper release tag for charmstore-client once the whitelist is updated?

cmars:
@marcoceppi Thanks :) we've decided to compile terms-client and plans-client in to charmstore-client, rather than build them here as plugins. That should improve the perceived performance of the CLI (doesn't have to enumerate over all the binaries and execute them for help, etc.) and simplifies the snap! Will push an update as soon as it's tagged & ready.

cmars:
@marcoceppi Ok, updated charmstore-client tag to 2.2.3.

PR #382, opened 2017-12-16
Purge duplicate wheels for module when overriding

### LINKED ISSUE #381: wheelhouse.txt is only additive
When building a charm, there is no way to actually override a dependency in `wheelhouse.txt`.  New entries instead simply add to the wheelhouse and at install time they are installed according to lexicographical ordering.  While this usually ends up with the desired result, it fails if you need to pin something to an older version and inflates the wheelhouse and thus the charm.

To address this, we will need to parse the wheelhouse entries at build time and ensure that we do the right thing.

gnuoy:
When the --wheelhouse-overrides option is used, duplicate wheels can
be produced in the resulting wheelhouse if the version pulled in by
the override wheelhouse is different from the version pulled in
during the regular build.

With this patch when the override wheelhouse is being processed it
removes any existing wheelhouses for the module being added. Since
the override wheelhouse is processed last the modules specified in
it are always the copies from the override.

Looking at wheel package format *1 hypens are replaced with
underscores in the distribution name and then hypens are used as
delimiters in the wheel file name *2:
{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform
tag}.whl.
Given that I think deleting wheels that match {distribution}-* is
safe.

*1 https://www.python.org/dev/peps/pep-0427/#escaping-and-unicode
*2 https://www.python.org/dev/peps/pep-0427/#file-name-convention

closes #381

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -769,6 +769,7 @@ class WheelhouseTactic(ExactMatch, Tactic):
         self.tracked = []
         self.previous = []
         self._venv = None
+        self.purge_wheels = False
 
     def __str__(self):
         directory = self.target.directory / 'wheelhouse'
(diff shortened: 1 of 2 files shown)


galgalesh:
I tested this locally and it works.

@johnsca, can you merge this please?

charmtools/build/tactics.py
```diff
@@ -769,6 +769,7 @@ def __init__(self, *args, **kwargs):
         self.tracked = []
         self.previous = []
         self._venv = None
+        self.purge_wheels = False
```
johnsca: I think we should actually make this default, but we can try it out with overrides and flip the switch later.

PR #389, opened 2018-01-29
Update display-name regex to match the juju/names package.

hatched:


mitechie:
Change looks ok but no tests with the updated regex?

mitechie:
👍 ty for the tests

hatched:
Thanks @mitechie

PR #395, opened 2018-05-01
Fix tactics defined in a layer not applying to that layer

### LINKED ISSUE #252: Tactics defined in layers do not execute against built artifact
## Checklist

## What version am I running?

```
master
```

I am using: Xenial

## Issue/Feature

When defining a tactic in a layer, that tactic only operates against that layer and not the entire built charm.

After breaking backwards compatibility with a layer I produced a tactic that will fail the build if the previous, broken, pattern is used.

### What I got

`charm build` succeeds despite having items that should be caught in the tactic.

```python
# tactics/ruby.py
from charmtools.build.tactics import ExactMatch, Tactic


class DependenciesTxtTactic(ExactMatch, Tactic):
    FILENAME = 'dependencies.txt'

    def __call__(self):
        raise Exception('dependencies.txt is deprecated see '
                        'https://github.com/battlemidget/juju-layer-ruby/pull/8')
```

```yaml
 # layer.yaml
includes: ['layer:basic']
tactics: ['tactics.ruby.DependenciesTxtTactic']
```

If I build a layer which depends on the modified `layer:ruby` with the above tactic the tactic is never executed.

johnsca:
Fixes #252

## Checklist

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -303,7 +303,8 @@ class Builder(object):
                 results["layers"].append(base_layer)
                 self.post_metrics('layer', base_layer.name, base_layer.fetched)
 
-    def build_tactics(self, entry, layer, next_config, output_files):
+    def build_tactics(self, entry, layer, next_config, current_config,
+                      output_files):
         relname = entry.relpath(layer.directory)
         existing_tactic = output_files.get(relname)
         tactic = Tactic.get(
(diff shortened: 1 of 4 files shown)


charmtools/build/builder.py
```diff
@@ -303,28 +303,32 @@ def fetch_dep(self, layer, results):
...
             layer,
             next_config,
+            current_config,
             existing_tactic,
         )
         output_files[relname] = tactic
 
     def plan_layers(self, layers, output_files):
+        current_config = BuildConfig()
         next_config = BuildConfig()
```
wallyworld: Instead of calling BuildConfig() twice, would it be more efficient to simply say:
next_config = current_config

PR #396, opened 2018-05-03
Make standard hooks generated from template

### LINKED ISSUE #98: Don't embed hooks
Currently, layered charms' hooks are an embedded chunk of Python. This might come from the base layer, or an old version of the base layer, or customized by an included layer, or customized by the main charm. This makes it difficult to propagate fixes, and is a bit of a mess.

Instead, lets not have any hooks in the hooks directory. List the hooks that each layer needs in its layers.yaml file. At composition time, copy hooks/hook.template into place for each hook and set executable permissions, if the hook does not already exist. 

This would allow changes in the base layer's hook.template to propagate. It would allow layers and charms to create a new hook.template if necessary. It would allow layers and charms to override the hook template just for those hooks that actually need it.

johnsca:
Fixes #98

## Checklist

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -357,6 +357,18 @@ class Builder(object):
         plan = [t for t in output_files.values() if t]
         return plan
 
+    def plan_hooks(self, layers, output_files, plan):
+        if self.HOOK_TEMPLATE_FILE not in output_files:
+            raise BuildError('At least one layer must provide %s',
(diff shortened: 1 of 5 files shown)


johnsca:
Codacy report is spurious (complaining that `self` isn't used in a test method)

johnsca:
@ajkavanagh @stub42 @galgalesh reviews requested as well.

PR #426, opened 2018-07-18
Add Py3 workaround to missed charm templates

### LINKED ISSUE #425: charm-create fails with "bytes not str" error
It seems that the switch to Python 3 missed some code in the charm create templates.

```
  File "/snap/charm/158/bin/charm-create", line 9, in <module>
    load_entry_point('charm-tools==2.2.4', 'console_scripts', 'charm-create')()
  File "/snap/charm/158/lib/python3.5/site-packages/charmtools/create.py", line 98, in main
    generator.create_charm()
  File "/snap/charm/158/lib/python3.5/site-packages/charmtools/generators/generator.py", line 93, in create_charm
    self.plugin.create_charm(user_config, tempdir)
  File "/snap/charm/158/lib/python3.5/site-packages/charmtools/templates/reactive_python/template.py", line 53, in create_charm
    self._template_file(config, path.join(root, outfile))
  File "/snap/charm/158/lib/python3.5/site-packages/charmtools/templates/reactive_python/template.py", line 64, in _template_file
    o.write(str(t))
  File "/snap/charm/158/usr/lib/python3.5/tempfile.py", line 622, in func_wrapper
    return func(*args, **kwargs)
TypeError: a bytes-like object is required, not 'str'
```

johnsca:
Fixes #425

### DIFF
--- a/charmtools/templates/python/template.py
+++ b/charmtools/templates/python/template.py
@@ -19,6 +19,7 @@ import logging
 import os
 import os.path as path
 import time
+import sys
 import shutil
 import subprocess
 import tempfile
@@ -64,7 +65,10 @@ class PythonCharmTemplate(CharmTemplate):
         o = tempfile.NamedTemporaryFile(
             dir=path.dirname(outfile), delete=False)
         os.chmod(o.name, mode)
-        o.write(str(t))
+        st = str(t)
+        if sys.version_info >= (3, ):
+            st = st.encode('UTF-8')
+        o.write(st)
         o.close()
         backupname = outfile + str(time.time())
         os.rename(outfile, backupname)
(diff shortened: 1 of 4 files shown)


charmtools/templates/python/template.py
```diff
@@ -64,7 +65,10 @@ def _template_file(self, config, outfile):
         o = tempfile.NamedTemporaryFile(
             dir=path.dirname(outfile), delete=False)
         os.chmod(o.name, mode)
-        o.write(str(t))
+        st = str(t)
+        if sys.version_info >= (3, ):
```
Cynerva: Is this check needed? Do we have to support both python2 and python3?

johnsca: Unfortunately, we do.  charm-tools is included directly via PyPI in some CI systems, though I don't know if `charm create` would actually be used by those.

Note, this was copied from one of the templates that [already included the check](https://github.com/juju/charm-tools/blob/master/charmtools/templates/bash/template.py#L64-L67).

PR #432, opened 2018-07-26
Add warning for switch from strict to classic confinement

johnsca:
This will need to be released to stable *just* prior to the actual confinement change.


johnsca:
Need to move `wrappers/`:

```
The snap/ directory is meant specifically for snapcraft, but it contains the following non-snapcraft-related paths, which is unsupported and will cause unexpected behavior:
- .gitignore
- wrappers
- wrappers/charm
```

helpers/snap-wrappers/charm
```diff
@@ -0,0 +1,28 @@
...
+    >&2 echo "$1"
+    if [ -t 1 ]; then
+        # reset if term
+        tput sgr0
+    fi
+}
+
+warn "To improve the developer experience, the confinement of this snap is changing "
+warn "from strict to classic. For security reasons, updates will be blocked until"
+warn "you manually refresh to authorize the new confinement."
```
kwmonroe: Can you add a 'howto' after this warning?  Something like:
```
warn "you manually refresh to authorize the new confinement:"
warn ""
warn "sudo snap refresh charm --classic"
```

PR #433, opened 2018-07-26
Switch to classic confinement

### LINKED ISSUE #415: Revert to classic confinement for charm snap
After some discussion, it was decided that since the charm command is basically a dev tool, it makes more sense for it to use classic confinement.  More, since the ELF patching feature was added to snapcraft, the issue that caused us to move to strict confinement (#339) should be resolved (though this will need testing).  This would fix #373.

One issue with moving back to classic confinement is that the update process will be silently broken for those who originally installed with strict confinement, since it (rightly so) won't update from strict to classic automatically.  The approach suggested to deal with this is to cut one more release with strict confinement that causes the command to print a deprecation warning encouraging the user to update manually.

johnsca:
Fixes #415

Process for merging this would be to merge #432 first, update the version for release, let it build and get to edge, then promote it to at least beta, all before merging this to master.

Note: This PR is currently against #432 rather than master so that the diff works out properly.

### DIFF
--- a/helpers/snap-wrappers/charm
+++ b/helpers/snap-wrappers/charm
@@ -7,24 +7,4 @@ export PREFIX=$SNAP
 export GIT_EXEC_PATH=$SNAP/usr/lib/git-core
 export LC_ALL=C.UTF-8
 
-warn() {
-    if [ -t 1 ]; then
-        # colorize if term
-        tput bold
-        tput setaf 3
-    fi
-    >&2 echo "$1"
-    if [ -t 1 ]; then
-        # reset if term
-        tput sgr0
-    fi
-}
-
-warn "To improve the developer experience, the confinement of this snap is changing "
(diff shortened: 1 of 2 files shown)


johnsca:
There are some checks and warnings about only operating under `$HOME` that should be removed in this PR as well, or it will give spurious complaints.

marcoceppi:
_(nit) this is like the 12th time we've changed confinement for this project. I hope it will be the last_

johnsca:
Second, but yes.  The main reason we moved away from classic confinement was because of leaks between the system Python and the snap-provided Python, which should be fixed now with the ELF patching, though this does need to be tested well.  The overall opinion is that dev tools simply don't make sense being strictly confined because they inherently require a level of access and trust that end-user apps do not.  E.g., snapcraft itself is and always will be classically confined, according to the maintainers.

PR #434, opened 2018-08-07
Updates for building k8s charms

johnsca:
Track charmstore-client master to get current changes into edge.
Also make proof respect some new fields for k8s type charms.


johnsca:
The new `auto-fetch` and `validator` keys are for the [docker-resource layer](https://github.com/juju-solutions/layer-docker-resource) (which I hope to fold into charms.reactive in a more general way) to provide the library with hints on how to handle resources and surface flags for them.

PR #435, opened 2018-08-09
Report detailed version for charmstore-client

johnsca:
Because the charmstore-client repo doesn't have official releases often, the base version number is often not indicative of what is actually included.  This extends the existing git_version support to include the charmstore-client component as well.

Example output:

```
charmstore-client 2.3.0+git-11-g5ea61cd
charm-tools 2.2.5+snap-x1+git-3-g32b98a2
```

### DIFF
--- a/charmtools/git_version.py
+++ b/charmtools/git_version.py
@@ -39,14 +39,10 @@ def _scm_version(repo_path):
 def format_version(version_info, ver_format):
     if ver_format == 'json':
         return json.dumps(version_info)
-    if 'SNAP_REVISION' in os.environ:
-        snap = '+snap-{}'.format(os.environ['SNAP_REVISION'])
-    else:
-        snap = ''
     pre_release = version_info.get('pre_release') or version_info.get('gitn')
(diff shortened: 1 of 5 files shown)


charmtools/git_version.py
```diff
@@ -88,3 +101,23 @@ def get_args(args=None):
...
+                             "info. Default uses long if it's a pre-release.")
+    parser.add_argument('--build-charm-version', action='store_true')
+    args = parser.parse_args(args)
+
+    return args
+
+
+if __name__ == '__main__':
+    args = get_args()
+    if args.build_charm_version:
```
Veebers: Personally I would find ```print(_build_charm_version())``` slightly clearer that this script outputs something that is consumed elsewhere (as opposed to _build_charm_version() writing a file or something else etc.)

snap/workarounds/charmstore-client-version
```diff
@@ -1 +1 @@
-2.3.0
+{"snap": "", "git": "", "gitn": 0, "version": "2.3.0", "pre_release": false}
```
johnsca: I could probably just delete this.

PR #436, opened 2018-08-20
made charm-build has the ability to parse devices metadata

ycliuhw:
Description of change
made `charm-build` has the ability to parse `devices` metadata.

### DIFF
--- a/charmtools/charms.py
+++ b/charmtools/charms.py
@@ -720,11 +720,11 @@ def validate_devices(charm, linter):
         messages will be written
 
     """
-    devices = charm.get('devices', None)
-    if devices is None:
+    devices = charm.get('devices', {})
+    if devices == {}:
         return
 
-    if not isinstance(devices, dict) or not devices:
+    if not isinstance(devices, dict):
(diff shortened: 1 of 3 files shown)


charmtools/charms.py
```diff
@@ -696,6 +712,33 @@ def validate_storage(charm, linter):
...
+    :param charm: dict of charm metadata parsed from metadata.yaml
+    :param linter: :class:`CharmLinter` object to which info/warning/error
+        messages will be written
+
+    """
+    devices = charm.get('devices', None)
+    if devices is None:
+        return
+
+    if not isinstance(devices, dict) or not devices:
```
johnsca: I don't think an empty set of devices should be an error. Typically, the section would be omitted entirely, but there could be cases where it gets included as an empty dictionary due to a build processes.  An empty dictionary should be treated the same as an omitted section.

ycliuhw: yes, you are right.

wallyworld:
[implement charm build tool device metadata parser](https://trello.com/c/2gslwBkq/342-implement-charm-build-tool-device-metadata-parser)

PR #445, opened 2018-09-14
Fix version causing build signature issues

johnsca:
The version file was getting added to the manifest even if it wasn't created, causing a modification error stating that it was removed after the last build.
    
It was also failing if the charm was a git repo that didn't have annotated tags instead of just using the sha.

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -955,6 +955,7 @@ class VersionTactic(Tactic):
         super(VersionTactic, self).__init__(
             charm / self.FILENAME, target, layer, next_config)
         self.charm = charm
+        self.wrote_sha = False
 
     def read(self):
         if self.entity.isfile():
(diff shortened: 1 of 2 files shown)


charmtools/build/tactics.py
```diff
@@ -973,7 +974,7 @@ def trigger(cls, entity, target, layer, next_config):
 
     def _try_to_get_current_sha(self):
         cmds = (
-            ('git', 'describe', '--dirty'),
+            ('git', 'describe', '--dirty', '--always'),
```
johnsca: For reference:

```
[johnsca@murdoch:~ … layers/vault/src] $ git describe --dirty
fatal: No names found, cannot describe anything.
[johnsca@murdoch:~ … layers/vault/src] $ git describe --dirty --always
f5ba548-dirty
```

PR #458, opened 2018-10-03
Fix version in setup.py during install

### LINKED ISSUE #457: charm-tools pypi package shows "unknown" version and has virtualenv build errors
Reproducer:

```
virtualenv venv
. venv/bin/activate
pip install charm-tools
pip freeze
```

johnsca:
During install, `setup.py` is called again and tries to pull the version info from git rather than using the prepared cached `VERSION` file.

Fixes #457

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -67,12 +67,10 @@ charm version --format=long
 
 PyPI should be kept up to date with full releases, but the authoritative source
 is always the snap.  Note: for the version to work on PyPI, it needs to be cached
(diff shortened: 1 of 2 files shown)


johnsca:
Tested with:

```
pip uninstall charm-tools
python setup.py sdist
pip install ./charm-tools-*.tar.gz
pip freeze | grep charm-tools
```

That gives "unknown" on master and a valid version on the branch.

Issue #438, opened 2018-08-29
Rename and improve --interface-service


johnsca:
Rename the `--interface-service` option to `--layer-index` to better match what it is.  Also, we should accept normal GitHub URLs, including forks and branches (e.g., https://github.com/Cynerva/layer-index/tree/gkk/test-subdir), and automatically convert them to the appropriate raw.githubusercontent.com form.

PR #460, opened 2018-10-10
Add support for local layer-index

johnsca:
The requests library doesn't support `file://` URLs, so we have to handle them specially.  This allows for a copy of the layer-index repo to be cloned locally and used instead.

Also rename `--interface-service` to `--layer-index`, partially addressing #438.


battlemidget:
LGTM

PR #462, opened 2018-10-12
Fix default value for --layer-index option

johnsca:
The implicit default of None for `--interface-service` was somehow overwriting the actual default for `--layer-index`.

Also fix the security complaint from GitHub with the older Paramiko version.


johnsca:
I'm not actually sure why paramiko is in setup.py.  That seems to be the only reference to it.

PR #465, opened 2018-10-16
Allow proof validation to be extended by layers

johnsca:
Adds support for layers to define new annotations on certain metadata section definitions (resources, storage, payloads, and devices) that are validated by proof and handled by the layer.

This also moves the layer-docker-resource specific fields (auto-fetch and validator) out of charm-tools proper and into the layer.

An example definition in a layer's `layer.yaml` would be:

```yaml
proof:
  resources:
    - name: auto-fetch
      type: Boolean
      missing: false
```

Which would allow the following annotation in a charm's `metadata.yaml`:

```yaml
name: mycharm
resources:
  tf-hub:
    type: oci-image
    auto-fetch: true
```

## Checklist


johnsca:
@wallyworld From your perspective specifically I wanted feedback on whether it's reasonable to allow charms to support fields that Juju doesn't know or care about.  I feel like this is a good compromise between being completely permissive about unknown fields like Juju vs being completely strict like charm proof currently, but I'd like your perspective on it as well.

Issue #468, opened 2018-10-25
charm version generator feature - builds fail if no vcs is present


ryan-beisner:
charm version generator feature - builds fail if no vcs is present
Feature implemented @ https://github.com/juju/charm-tools/pull/437

I think we should not require vcs in order to build.

Reproducer:

- Git clone a charm, cd to it.
- ```rm -rf .git```
- build the charm
- error:

```
utils: process: git describe --dirty --always (128)
utils: bzr: ERROR: Not a branch: "/home/ubuntu/git/charm-aodh/src/".

utils: process: bzr version-info (3)
Traceback (most recent call last):
  File "/home/ubuntu/git/charm-aodh/.tox/build/bin/charm-build", line 11, in <module>
    sys.exit(main())
  File "/home/ubuntu/git/charm-aodh/.tox/build/local/lib/python2.7/site-packages/charmtools/build/builder.py", line 834, in main
    build()
  File "/home/ubuntu/git/charm-aodh/.tox/build/local/lib/python2.7/site-packages/charmtools/build/builder.py", line 636, in __call__
    self.generate()
  File "/home/ubuntu/git/charm-aodh/.tox/build/local/lib/python2.7/site-packages/charmtools/build/builder.py", line 589, in generate
    self.exec_plan(self.plan, self.layers)
  File "/home/ubuntu/git/charm-aodh/.tox/build/local/lib/python2.7/site-packages/charmtools/build/builder.py", line 562, in exec_plan
    tactic()
  File "/home/ubuntu/git/charm-aodh/.tox/build/local/lib/python2.7/site-packages/charmtools/build/tactics.py", line 999, in __call__
    new_sha = self._try_to_get_current_sha()
  File "/home/ubuntu/git/charm-aodh/.tox/build/local/lib/python2.7/site-packages/charmtools/build/tactics.py", line 992, in _try_to_get_current_sha
    except FileNotFoundError as e:
NameError: global name 'FileNotFoundError' is not defined
ERROR: InvocationError: '/home/ubuntu/git/charm-aodh/.tox/build/bin/charm-build --log-level DEBUG -o /home/ubuntu/git/charm-aodh/build src'
___________________________________ summary ____________________________________
ERROR:   build: commands failed
```

ryan-beisner:
```
(build) ubuntu@beisner-bastion:~/git/charm-aodh$ pip freeze | grep charm
snap charm-tools==2.4.5
libcharmstore==0.0.8
[...]

PR #469, opened 2018-10-25
Update exception to work with py2 and py3

ryan-beisner:
Closes: #468

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -989,7 +989,7 @@ class VersionTactic(Tactic):
                     sha = utils.Process(cmd)()
                     if sha:
                         return sha.output
-                except FileNotFoundError as e:
+                except IOError as e:
                     log.debug(e)
                     continue
         return ""


charmtools/build/tactics.py
```diff
@@ -989,7 +989,7 @@ def _try_to_get_current_sha(self):
                     sha = utils.Process(cmd)()
                     if sha:
                         return sha.output
-                except FileNotFoundError as e:
+                except IOError as e:
                     log.debug(e)
```
stub42: What you have will ignore diskfull and other exceptions we don't want to catch.
```suggestion
                    if e.errno != errno.ENOENT:
                        raise
                    log.debug(e)
```

johnsca: +1

johnsca:
Closing in favor of #479 with @stub42's suggestion, plus a test, plus fixing the exception type (has to be `OSError` rather than `IOError`).

PR #473, opened 2019-01-18
Add revision info to build manifest

### LINKED ISSUE #403: Add info about layer sha / rev to .build.manifest
`charm-build` currently creates a `.build.manifest` file with info about what was built.  It would be good to include in that the specific rev or sha for each layer that was built, so that it can be audited exactly what went into building the charm.

johnsca:
This allows tracking of exactly what a given charm was built with.
The build manifest is now (as of the latest charm snap release) uploaded to the store.

This doesn't provide any way of using this info to manage a rebuild, but it can at least be done manually via local checkouts of the layers, if needed.

Fixes #403

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -27,6 +27,7 @@ from charmtools.build.fetchers import (
     LayerFetcher,
     get_fetcher,
     FetchError,
+    RepoFetcher,
 )
 from charmtools.version import cached_charm_tools_version, format_version
 
@@ -65,6 +66,7 @@ class Fetched(Configable):
         self.directory = None
         self._name = name
         self.fetched = False
+        self.revision = None
 
     @property
     def name(self):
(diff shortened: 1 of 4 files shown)


Cynerva:
It looks like this doesn't work for layers that are in subdirs. For example, when building kubernetes-master:

```
{
  "layers": [
    {
      "rev": "fcdcea4e5de3e1556c24e6704607862d0ba00a56",
      "url": "layer:options"
    },
    {
      "rev": "baf582236a7b946da4cd140fd6532a2e0c2774d5",
      "url": "layer:basic"
    },
    ...
    {
      "rev": "",
      "url": "layer:kubernetes-common"
    },
    {
      "rev": "",
      "url": "kubernetes-master"
    },
```

Any idea what's going on there?

johnsca:
Yes, that would be because the logic for `get_revision` doesn't handle it properly.  Fix incoming.

Cynerva:
The fix helped, but I still see an issue when building with `--no-local-layers`:
```
{
  "rev": null,
  "url": "layer:kubernetes-common"
},
```

I suspect the subdir handling fetch code ([link](https://github.com/juju/charm-tools/blob/758045c7a451cd136c7b5713068d6a263b21ebb2/charmtools/build/fetchers.py#L129)) pulls layer-kubernetes-common out of its git repo, which prevents get_revision from working. Does that sound right?

Could it do a symlink instead of a rename, or something?

PR #474, opened 2019-01-24
Improve error handling in TERM detection

### LINKED ISSUE #472: Running in tmux or screen with TERM="screen-256color" causes charm-build to fail
With charm-build, installed via snap (2.4.5+git-1-gd62c072, snap relase 298) the same bug described in #344 is observable when running charm-build inside of `tmux`.

This bug is reproducible by setting `TERM="screen-256color"`.

To work around this bug, I am setting `TERM=linux` in the Makefile I am using to drive my charm build.

johnsca:
Handle errors better when detecting the `TERM` type for `charm-build`.

Also add a `--force-color` option for `charm-build` and rename `--force-raw` for `charm-inspect`

Fixes #472

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 import argparse
 import blessings
+import curses
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


johnsca:
This is available from the snap store in branch edge/bug-472 for testing.

sparkiegeek:
:+1: LGTM, works on my 18.10 box with `TERM=screen-256color` which previously failed

PR #481, opened 2019-02-06
Fix typo to make the argument alias keep working

tai271828:
This commit fix the typo of the previous commit (5807aa1d) so we could
retain the alias of `--build-dir` when using `--output-dir`.

## Checklist

 - [ ] Have you followed [Juju Solutions hacking guide?](https://hacking.juju.solutions)
    - The link does not work anymore. Both of the links are empty as well https://github.com/juju-solutions/hacking.juju.solutions and https://github.com/juju-solutions/hacking-guide
 - [X] Are all your commits [logically] grouped and squashed appropriately?
 - [X] Does this patch have code coverage?
 - [X] Does your code pass `make test`?

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -660,7 +660,7 @@ class Builder(object):
         if not self.build_dir:
             if self.output_dir:
                 self.build_dir = self.output_dir / series
-            if charm_build_dir:
+            elif charm_build_dir:
                 self.build_dir = path(charm_build_dir)
             elif juju_repo_dir:
                 self.build_dir = path(juju_repo_dir) / series


johnsca:
The removal of the `dest` param was intentional, to allow `--output-dir` to be handled [separately and slightly differently](https://github.com/juju/charm-tools/commit/5807aa1d594f4e52616bb2ecc3e5c0e23d50f475#diff-3307db7779607c17be6e529e2873b25cR662) in order to retain previous semantics where the `--output-dir` param behaved more like `JUJU_REPOSITORY` than `CHARM_BUILD_DIR`.

Is it not working for you as expected in the same way that it did previously?

tai271828:
Hi, johnsca. Thanks for the clear elaboration. I updated the pull request to fit the use case as what you have described. They are tested by `make deps; make; make test`, used to build my charm, and works like a _charm_.

As the original unexpected behavior of `--output-dir` for me (charm tool snap version 318), `--output-dir` did not work at all no matter if `JUJU_REPOSITORY` or `CHARM_BUILD_DIR` were assigned or not assigned. The option `--output-dir` is totally ignored to create the corresponding build folder when charm-build my layers.

My pull request is ready to review again. I will appreciate your comments. Thanks.

johnsca:
2.5.2 should show up with this in edge soon, at which point I'll promote it through stable.

PR #484, opened 2019-02-11
Address security alerts from GitHub

johnsca:
Update colander, requests, and PyYAML.

Technically, the fixed version of PyYAML isn't released yet, so this might not actually appease GitHub, but this also ensures we are explicitly using safe_load everywhere except the one place we explicit want to be a little unsafe.


tox.ini
```diff
@@ -4,8 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-#envlist = py27, py32, py34, py35
-envlist = py27, py35
+envlist = py27, py3
```
thedac: We have been having this debate ourselves. While the py3 target is easier it is also useful to assert your support py34, py35, py36, py37 or whichever ones you believe should work. 

You might consider adding these additional targets.

PR #489, opened 2019-02-20
Add basic documentation

johnsca:
This mostly just documents the current set of available commands and how to get further details using the CLI, as well as information about the build tactics, including how to create a custom tactic and the info about the base Tactic class.

Docs can be viewed here: https://charm-tools.readthedocs.io/en/johnsca-docs/


docs/_extensions/automembersummary.py
```diff
@@ -0,0 +1,91 @@
...
+
+import inspect
+
+from docutils.parsers.rst import directives
+from sphinx.ext.autosummary import Autosummary
+from sphinx.ext.autosummary import get_import_prefixes_from_env
+from sphinx.ext.autosummary import import_by_name
+
+
+class AutoMemberSummary(Autosummary):
```
johnsca: This has been cargo-culted into a few repos now.  I really should extract this and put it on pypi.

charmtools/build/tactics.py
```diff
@@ -170,18 +227,15 @@ class IgnoreTactic(Tactic):
     """
     @classmethod
     def trigger(cls, entity, target, layer, next_config):
-        """
-        Match if the given entity will be ignored by the next layer.
-        """
+        ""  # suppress inherited doc
```
kwmonroe: Hello future traveller.  If you're maintaining this code and stumble upon this PR on a quest to find out why the heck there's an empty string followed by an inline comment, you're just like me!

TIL that *no* docstring will inherit the docstring from base class methods. To prevent this, @johnsca has cleverly added empty docstrings where the inheritance is not wanted.

PR #491, opened 2019-02-22
Add promulgate and unpromulgate commands

johnsca:
This adds promulgate and unpromulgate commands to the charm snap.  These commands do some basic integrity checking to ensure that things don't get promulgated into a bad state, including:

  * Ensuring that there isn't already a charm promulgated with that name
  * Ensuring that the charm is released to stable
  * Ensuring that the stable branch is readable by "everyone"
  * After promulgation, ensuring that the owner can release updates to     their charm

See https://discourse.jujucharms.com/t/promulgate-a-charm-for-charmers-team/1206

This also updates the snap build to use core18, which is required for bhttp.


johnsca:
Note: These commands use some Python3 only features that probably won't work if run from a pip installed version (especially Python2).  However, the only reason we really support that usage still is for OpenStack automation, which shouldn't be doing promulgation anyway.

PR #492, opened 2019-02-22
Fix and improve charm-layers

johnsca:
The `charm layers` command was broken with the addition of revisions to the manifest.  This fixes that, and also adds some additional functionality, such as the ability to annotate the source layer when colors aren't available, and a larger number of colors to choose from so that charms with many layers won't run out quite as soon.

layer-inspect

### DIFF
--- a/charmtools/build/inspector.py
+++ b/charmtools/build/inspector.py
@@ -6,18 +6,18 @@ from charmtools import utils
 
 theme = {
     -1: "bright_black",
-    0: "white",
+    0: "normal",
     1: "green",
     2: "cyan",
     3: "magenta",
     4: "yellow",
     5: "red",
-    6: "bright_white",
-    7: "bright_green",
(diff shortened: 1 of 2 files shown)


charmtools/build/inspector.py
```diff
@@ -5,12 +5,19 @@
 from charmtools import utils
 
 theme = {
-    0: "normal",
```
kwmonroe: there's a few references to `theme.get(x, 'normal')` below.  I think you'll need to either `s/white/normal` here or update those refs below.

johnsca: I think those are correct.  `normal` just means "go back to default styling" which might not actually be `white` depending on the user's term config.

johnsca: Ah, but to your point, if a light theme, explicitly using `white` won't render well.

PR #508, opened 2019-03-11
Refactor version handling in snap to work with core18

johnsca:
Switching to core18 highlighted some flaws in how we were processing the wrappers and versions, and also caused us to hit issues with the use of `stage` and `prime` where any explicit includes cause all default excludes to be dropped.  These led to the snap build being broken.


johnsca:
See https://forum.snapcraft.io/t/python-classic-snap-confusion-frustration-core18-entry-point-scripts-organize-stage-prime-override-build/10317 for extended context.

PR #514, opened 2019-04-02
Add --layer-index arg to pull-source

tvansteenburgh:
Provides a way to override the default layer-index used by the `charm pull-source` command.


tvansteenburgh:
This will be superseded by a different PR, incoming.

PR #515, opened 2019-04-02
Add support for layer-index to pull-source and multiple indexes

johnsca:
Adds support for `--layer-index` and `--fallback-layer-index` option to `charm pull-source` and the support for the latter to `charm build` as well.

Also fix an infinite recursion bug when dealing with local interface layers, due to an incorrect `super()` call.

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -883,7 +883,12 @@ def main(args=None):
     parser.add_argument('--interface-service',
                         help="Deprecated: use --layer-index")
     parser.add_argument('--layer-index',
-                        default=LayerFetcher.LAYER_INDEX)
+                        help='URL of main index to use to look up layers '
+                             '(default: {})'.format(
+                                 LayerFetcher.LAYER_INDEXES[0]))
+    parser.add_argument('--fallback-layer-index',
+                        help='URL of index to use to look up layers '
+                             'not found in main layer index')
     parser.add_argument('--no-local-layers', action="store_true",
                         help="Don't use local layers when building. "
                         "Forces included layers to be downloaded "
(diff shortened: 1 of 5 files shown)


charmtools/build/builder.py
```diff
@@ -882,8 +882,14 @@ def main(args=None):
                         action="store_true")
     parser.add_argument('--interface-service',
                         help="Deprecated: use --layer-index")
-    parser.add_argument('--layer-index',
-                        default=LayerFetcher.LAYER_INDEX)
+    parser.add_argument('-i', '--layer-index',
+                        help='One or more index URLs use to look up layers, '
+                             'separated by commas. Can include the token '
+                             'DEFAULT, which will be replaced by the default '
+                             'index{}: {}'.format(
```
tvansteenburgh: It would be nice to 1) show an example of what this looks like (e.g. https://mysite.com/my-index/,DEFAULT), and also show what the default index is (same for the help on pullsource.py).

tvansteenburgh: I changed my mind about this, I think what's here is fine.

charmtools/pullsource.py
```diff
@@ -211,6 +222,14 @@ def setup_parser():
         'dir', nargs='?',
         help='Directory in which to place the downloaded source.',
     )
+    parser.add_argument(
+        '-i', '--layer-index',
+        help='One or more index URLs use to look up layers, '
```
tvansteenburgh: used

charmtools/build/builder.py
johnsca: The default value is shown (the join bit).  I'll add the example.

charmtools/build/fetchers.py
```diff
@@ -34,14 +34,33 @@ def can_fetch(cls, url):
...
+    @classmethod
+    def set_layer_indexes(cls, layer_indexes):
+        if not layer_indexes:
+            return
+        if isinstance(layer_indexes, str):
+            layer_indexes = layer_indexes.split(',')
+        new_indexes = []
+        for layer_index in layer_indexes:
+            if layer_index == 'DEFAULT':
+                new_indexes.extend(cls.LAYER_INDEXES)
```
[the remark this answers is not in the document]
johnsca: It should be equivalent, but the other probably makes more sense.

PR #523, opened 2019-06-02
Upgrade to use py3.7

chenrui333:
Description of change

Relates to https://github.com/Homebrew/homebrew-core/pull/40496

Trying to upgrade to use Python 3.7 (which travis has the support with `dist: xenial`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
-sudo: false
 language: python
+dist: xenial
 python:
   - "2.7"
-  - "3.5"
+  - "3.7"
 install: pip install tox-travis
 script: tox
--- a/tox.ini
+++ b/tox.ini
@@ -4,8 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-#envlist = py27, py32, py34, py35
-envlist = py27, py35
+envlist = py27, py37
 skipsdist = true
 
 [testenv]


chenrui333:
Build passed, ref: https://travis-ci.com/chenrui333/charm-tools/builds/113995640

.travis.yml
```diff
@@ -1,7 +1,7 @@
-sudo: false
```
chenrui333: Travis side has consolidated this Linux build infra.

johnsca:
GH won't let me merge this without the Travis result.  I can't find in the repo settings how to allow me to bypass that.

chenrui333:
I see, I think I saw something similar before.

Usually restart the build would solve the issue.

johnsca:
It's not even showing up in the build history or PR list on Travis, which I think is the issue.  I'm going to close and re-open the PR to force it to trigger again.

chenrui333:
give me one sec.

chenrui333:
watch out on this new build: https://travis-ci.org/juju/charm-tools/builds/541784697

johnsca:
There was also a config issue on the Travis side.  But it looks like it's going now.  👍

chenrui333:
yeah, had the same issue before 😄 

Glad it got sorted out.

PR #520, opened 2019-05-01
Fix installing from git without vergit installed

johnsca:
Another corner case is installing directly from git but without vergit available, in which case there's no way to determine the version and we have to fall back to `None` (a.k.a. `0.0.0`).

Fixes #518

PR #527, opened 2019-06-05
Resolve vergit python package dependency

ryan-beisner:
Closes: 526


johnsca:
vergit is not intended to be an installation-time dependency.  This should have been addressed by #519 and #520 so it seems like there's another corner case that needs to be addressed.

PR #540, opened 2019-08-09
Cleanup cached layers / interfaces after build

### LINKED ISSUE #539: Charm build leaving temp files behind
## What version am I running?

I ran the following command: `snap info charm` and got the following ouput:

```
name:      charm
summary:   charm and charm-tools
publisher: Cory Johns (johnsca)
contact:   juju@lists.ubuntu.com
license:   unset
description: |
  charmstore-client and charm-tools
commands:
  - charm
snap-id:      2Rryoc2ylScfbFl4eQtpntHD9iuZuMvt
tracking:     edge
refresh-date: 14 days ago, at 14:19 CDT
channels:
  stable:    2.6.1               2019-07-12 (357) 58MB classic
  candidate: 2.6.1               2019-07-12 (357) 58MB classic
  beta:      2.6.1               2019-07-12 (357) 58MB classic
  edge:      2.6.1+git-1-28f7a67 2019-07-17 (359) 58MB classic
installed:   2.6.1+git-1-28f7a67            (359) 58MB classic
```

I am using: Ubuntu 18.04

I expected charm build to clean up after itself

### What I got

I got remnants of old builds sitting in `~/.cache/charms/`, see this output:

```
$ du -sh ~/.cache/charms/
2.3G	~/.cache/charms/
```

```
 du -sh ~/.cache/charms/* | sort -hr | head -n 5
165M	~/.cache/charms/pipelines-api
156M	~/.cache/charms/pipelines-ui
146M	~/.cache/charms/ambassador
143M	~/.cache/charms/mariadb-k8s
142M	~/.cache/charms/minio
```

```
ls ~/.cache/charms/pipelines-api/interface/ | head -n 10
http
mysql
tmp00y5kvrf
tmp0beow9ef
tmp0ra9_dk_
tmp0uckszle
tmp0y7jn120
tmp0yig2tas
tmp12cyvae9
tmp14bblbff
```

There are lots of the `tmp*` folders in both `interface/` and `layer/` under each charm directory. I would hazard a guess that this is due to me building bundles named e.g. `/tmp/foo123.yml`, due to the fact that I have a build process that builds each charm in the `bundle.yaml`, and then creates a temp bundle containing a path to the built charm for deployments.

johnsca:
The build process currently leaves directories under `~/.cache/charm/{layer,interface}/` which over time can build up and waste space.

Fixes #539

Additionally:

  * Fix handling of excludes in interface layers. The `ignore` setting is honored by interface layers but actually uses the mechanic of `exclude`, while `exclude` itself is not honored. This adds support for `exclude` with the proper behavior and leaves `ignore` as-is for compatibility.

  * Improve error reporting during interface layer linting when a file fails to decode.

  * Fix a snapcraft error if python binary already exists.

### DIFF
--- a/snap/hooks/post-refresh
+++ b/snap/hooks/post-refresh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# clean up orphaned cache directories which take up unused space
+for user in /home/*; do
+    if [ -e $user/.cache/charm ]; then
+        rm -rf $user/.cache/charm/*
+    fi
+done
--- a/snap/hooks/post-refresh
+++ b/snap/hooks/post-refresh
@@ -1,8 +1,4 @@
 #!/bin/bash
 
 # clean up orphaned cache directories which take up unused space
-for user in /home/*; do
-    if [ -e $user/.cache/charm ]; then
-        rm -rf $user/.cache/charm/*
-    fi
-done
+rm -rf /home/*/.cache/charm/*
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -48,7 +48,7 @@ parts:
       vergit --format=json > $SNAPCRAFT_PART_INSTALL/charm-tools-version
       snapcraftctl set-version $(vergit)
       # ensure we don't accidentally pick up system py2
-      ln -s python3 $SNAPCRAFT_PART_INSTALL/usr/bin/python
+      ln -s python3 $SNAPCRAFT_PART_INSTALL/usr/bin/python || true
       # copy our wrappers into the right place; note: we cannot use organize,
       # because it will break things (see forum link below)
       cp -R helpers/snap-wrappers $SNAPCRAFT_PART_INSTALL/bin/wrappers
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -647,6 +647,7 @@ class Builder(object):
         self.validate()
         self.find_or_create_target()
         self.generate()
+        self.cleanup()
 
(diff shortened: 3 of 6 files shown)


snap/hooks/post-refresh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# clean up orphaned cache directories which take up unused space
+for user in /home/*; do
```
knkski: Could this just be `rm -rf /home/*/.cache/charm/*`?

snap/snapcraft.yaml
```diff
@@ -48,7 +48,7 @@ parts:
       vergit --format=json > $SNAPCRAFT_PART_INSTALL/charm-tools-version
       snapcraftctl set-version $(vergit)
       # ensure we don't accidentally pick up system py2
-      ln -s python3 $SNAPCRAFT_PART_INSTALL/usr/bin/python
+      ln -s python3 $SNAPCRAFT_PART_INSTALL/usr/bin/python || true
```
hyperbolic2346: || true to prevent issue of file exists? why not -f?

johnsca: Yeah, it was to fix an issue I hit building the snap locally, which I assumed was due to a change in snapcraft but might have been a slightly dirty build VM.  I used `|| true` instead of `-f` because I didn't think of `-f`.  :p

hyperbolic2346: There is also the potential advantage of not smashing someone's custom link for some reason.

PR #543, opened 2019-08-29
Rename README files with markdown extension

timClicks:
README files within charm templates use the .ex extension. This confuses IDEs and other systems that depend on the file extension to render them correctly. 

## Checklist


jameinel:
I thought the point of the .ex was that it was an example and you are
expected to rename it as part of making it your own.

It maybe be easier for people for them to start as .md, and just use the
proof check to make sure someone puts something meaningful in there?

John
=:->

timClicks:
The `charm proof` tool does include a check to see that your readme doesn't contain the boilerplate.

PR #554, opened 2019-10-08
Allow boolean config options to have null default

johnsca:
It is occasionally useful to have an explicit null / None default value for boolean config options to detect the difference between the user explicitly wanting false or not. This changes the proof message to informational.

### DIFF
--- a/charmtools/charms.py
+++ b/charmtools/charms.py
@@ -68,7 +68,7 @@ KNOWN_OPTION_TYPES = {
     'boolean': bool,
 }
 
-ALLOW_NONE_DEFAULT = (six.string_types[0], int, float)
+ALLOW_NONE_DEFAULT = (six.string_types[0], int, float, bool)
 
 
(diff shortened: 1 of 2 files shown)


johnsca:
Test results:

```
[johnsca@raza-u:~ … ld/openstack-integrator] lxd:admin/default:5 $ charm proof .
I: config.yaml: option bs-version has no default value
W: config.yaml: option trust-device-path has no default value
W: config.yaml: option ignore-volume-az has no default value
[johnsca@raza-u:~ … ld/openstack-integrator] lxd:admin/default:5 100 $ ~/juju/tools/charm-tools/.tox/py36/bin/charm-proof .
I: config.yaml: option bs-version has no default value
I: config.yaml: option trust-device-path has no default value
I: config.yaml: option ignore-volume-az has no default value
```

battlemidget:
LGTM

PR #562, opened 2020-03-25
Tweak Build Snap action

johnsca:
* Rename output artifact to `charm.snap`; this will download as `charm.snap.zip` which will need to be unzipped to get the actual snap
* Rename yaml file so we can add other actions in the future.


johnsca:
Hrm.  I don't think the Build Snap action change will take effect on this PR.

PR #564, opened 2020-04-17
Make output_dir the same as build_dir

sabaini:
Per help text, output_dir and build_dir should be treated the same,
making it so

Also see juju#563


johnsca (approved):
Apologies for the long-overdue merge, but thanks for this fix!

PR #568, opened 2020-06-06
Added k8s-operator template

chris-sanders:
Add a template k8s-operator which sets up an operator charm with k8s specific configuration. Includes harness for unit testing and zaza for functional testing.

### DIFF
--- a/charmtools/templates/k8s_operator/__init__.py
+++ b/charmtools/templates/k8s_operator/__init__.py
@@ -1,0 +1,19 @@
+#!/usr/bin/env python
+#
+#    Copyright (C) 2010  Canonical Ltd.
+#    Author: James Hebden <james.hebden@canonical.com> 
+#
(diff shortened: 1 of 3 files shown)


johnsca:
I would prefer to see new template additions use the approach taken if #558 which is to just add a requirements line to the repo and have the repo include a `setup.py` which uses the `charmtools.templates` entry point to register its templates.  This will minimize the code living in charm-tools and allow you to iterate on the template class itself without being gated here.

chris-sanders:
Sure that looks like a very clean pattern, I'll see when I can find time to rework this in that fashion.

johnsca:
Per #578 we want to deprecate creating new charms with the `charm` snap and instead move to `charmcraft init` going forward, so I'm inclined to just close this instead, if there are no objections.

chris-sanders:
That's fine I'll close this

PR #569, opened 2020-06-23
Process wheelhouse.txt holistically rather than per-layer

johnsca:
Processing the wheelhouse.txt file from each layer in isolation means that pip can't resolve conflicts, duplicates, or dependency version pinning which spans across layers.  This can lead to duplicate versions of libraries being installed which violate constraints to avoid certain upstream bugs.

This came up in both of:

* https://bugs.launchpad.net/charm-flannel/+bug/1884598
* https://bugs.launchpad.net/charm-octavia/+bug/1884164

And in both cases trying to resolve it on a per-layer or per-charm basis results in a whack-a-mole process of catching all of the layers which indirectly reintroduce the unpinned version of a 
dependency pinned in a base layer.

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -914,7 +914,8 @@ def main(args=None):
                         help="Provide a wheelhouse.txt file with overrides "
                              "for the built wheelhouse")
     parser.add_argument('-W', '--wheelhouse-per-layer', action="store_true",
-                        help="Use old per-layer wheelhouse processing")
+                        help="Deprecated: Use original wheelhouse processing "
+                             "method (see PR juju/charm-tools#569)")
     parser.add_argument('-v', '--verbose', action='store_true', default=False,
                         help="Increase output (same as -l DEBUG)")
     parser.add_argument('--debug', action='store_true',
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -148,6 +148,7 @@ class Builder(object):
         self._top_layer = None
         self.hide_metrics = os.environ.get('CHARM_HIDE_METRICS', False)
         self.wheelhouse_overrides = None
+        self.wheelhouse_per_layer = False
         self._warned_home = False
 
     @property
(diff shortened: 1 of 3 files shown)


charmtools/build/builder.py
```diff
@@ -904,11 +905,17 @@ def main(args=None):
                         "from the interface service.")
     parser.add_argument('-n', '--name',
                         help="Build a charm of 'name' from 'charm'")
-    parser.add_argument('-r', '--report', action="store_true",
+    parser.add_argument('-r', '--report', action="store_true", default=True,
                         help="Show post-build report of changes")
+    parser.add_argument('-R', '--no-report', action="store_false",
```
kwmonroe, line 910: -r and -R seem to be contradictory; how about we kill the -R codepath and just use -r=True|False instead?

johnsca: I tried to figure out a way to implement the `-r[=true/|false]` pattern but the case that always breaks is the current behavior that many may be using:

```
$ charm build -r ./src
```

There's no way with argparse to say "consume an arg, but only if it matches one of these choices".  It will either not consume any args, always consume an arg, or try to consume an arg but blow up if it doesn't match one of the choices.

Additionally, I found at least one example with [`git -p` and `git -P`](https://git-scm.com/docs/git#Documentation/git.txt--p) which have the same contradictory meaning, and I know I've seen other cases and not found it terribly confusing.  In general, and in the case of this change, the behavior is what I would expect: either of the options can be given, even multiple times, and the last one present on the command line takes precedence.

kwmonroe: @johnsca i appreciate the attempt.  I didn't intend for you to bend over backwards, but it sure would have been nice if "honor a bool value" would *just work* for argparse optargs.  Anywho, I'm annoyed but not blocking contradictory flags; as you said, last in gets the cake, and i think that's a reasonable UX that charm builders can shoulder.

All that said, given that `--report` is now the default, I would ask you to consider #570.  That is, deprecate `-r` with some stderr message that says `-r is default and therefore deprecated; use -R to prevent charm build reports`.

johnsca:
This has been released in the snap and to PyPI with 2.7.5

PR #581, opened 2020-08-26
Add --branch option to pull-source

### LINKED ISSUE #580: pull-source needs a --branch option
When used with layers which use the `subdir` field in the index, `charm pull-source` loses the SCM metadata when it copies out the subdirectory.  This prevents the ability to switch to a specific branch.  Adding a `--branch` option to switch branches before copying out the subdirectory would allow for this.

johnsca:
Adds the ability to specify a branch to switch to when fetching the source of a charm or layer.  This is particularly useful for layer
entries which use the `subdir` field, since that can lose the SCM metadata and prevent post-pull switching.

Fixes #580
Fixes [lp:1893081](https://bugs.launchpad.net/charmed-kubernetes-testing/+bug/1893081)


johnsca:
Drive-by: switched from nose to pytest

PR #585, opened 2020-10-05
Add reproducible charm build feature

ajkavanagh:
This feature adds some command line options to allow the generation of
the build.lock file in the top layer.  This can then be used to rebuild
the charm with the exact same set of layers and python modules. Also the
ability to track a specific branch is also enabled.

Description of change

## Checklist

 - [ ] Are all your commits [logically] grouped and squashed appropriately?

### DIFF
--- a/docs/reproducible-charms.rst
+++ b/docs/reproducible-charms.rst
@@ -65,7 +65,8 @@ If a lock file (``build.lock``) is available in the top layer, then it will be
 used to control the versions of the layers and modules *by default*.  i.e. the
 presence of the lock file controls the build.
 
-Two options are available to control the build when a lock file is present:
+Three options are available which can influence the build when a lock file is
+present:
 
  * ``--ignore-lock-file``
  * ``--use-lock-file-branches``
--- a/docs/reproducible-charms.rst
+++ b/docs/reproducible-charms.rst
@@ -1,0 +1,101 @@
+Reproducible Charms
+===================
+
+When building charms, multiple layers are brought together in an ordered,
+depth-first recursive fashion.  The individual files of each layer are merged,
+and then python modules are brought in according to ``wheelhouse.txt`` files
+that may exist in each layer.
+
+Layers (and Interfaces) are typically Git repositories, and by default the
+default branch (usually called ``master``) of the repository is fetched and
+used.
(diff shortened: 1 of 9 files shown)


ajkavanagh:
Note, new feature still requires some tests.

docs/reproducible-charms.rst
```diff
@@ -0,0 +1,101 @@
...
+The name of the lock file is ``build.lock``.
+
+Rebuilding the charm from the lock file
+---------------------------------------
+
+If a lock file (``build.lock``) is available in the top layer, then it will be
+used to control the versions of the layers and modules *by default*.  i.e. the
+presence of the lock file controls the build.
+
+Two options are available to control the build when a lock file is present:
```
johnsca, line 68:
```suggestion
Three options are available which can influence the build when a lock file is present:
```

johnsca (commented):
This is definitely a long-needed feature and the usage pattern and implementation look good to me.

Adding tests for this does seem like it will be different because it seems like to do it properly, you'd need to create temporary layers and interfaces with version control and commit(s), do a build, then add more commit(s) and do another build to validate the lock is honored.  And that doesn't even cover the locking in the wheelhouse; maybe that could be managed with a file-system-based [self-hosted simple respository](https://packaging.python.org/guides/hosting-your-own-index/) that the test creates and manages?

Given the complexity of adding an in-repo test for this, the fact that this style of charm building is mostly deprecated in favor of charmcraft (at least for new charms), and that I expect that this will immediately and extensively used in the OpenStack charms' build and CI process (if it hasn't already) which we already manually gate releases of this tool on anyway, maybe it would be sufficient to rely on that process to validate this functionality?

docs/reproducible-charms.rst
ajkavanagh: I'll revise the doc as per the suggestion above.  The 'good' news with the patch is that it only comes into being if the relevant lock file is present; otherwise it has no effect.  So this should mitigate any issues around the extra feature.  Also, I've ensured that the existing tests still pass, so there should be no regressions.  I'll do the fix, update, and drop the WIP.  thanks!

johnsca (approved):

PR #587, opened 2020-11-10
fix snap build for updated charmstore-client

mhilton:
Description of change

## Checklist


johnsca (approved):
Seems pretty sus, but I'll vote to keep.  😉  Thanks for the fix.

PR #588, opened 2020-12-02
Fix TypeError from linter on X.Y min-juju-version

johnsca:
Because YAML can end up treating a bare X.Y number as a float rather than a string, the linter was blowing up with an exception
(`TypeError: expected string or bytes-like object`) instead of a useful error about the format of the min-juju-version field.

Drive-by: Update to match [current regexp from upstream][regexp].

[regexp]: https://github.com/juju/version/blob/d414598/version.go#L108

### DIFF
--- a/charmtools/charms.py
+++ b/charmtools/charms.py
@@ -800,8 +800,8 @@ def validate_min_juju_version(charm, linter):
     if 'min-juju-version' not in charm:
         return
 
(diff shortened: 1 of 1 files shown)


knkski (approved):
Specifying `{1-9}` instead of `+` is kind of a weird choice, but since it's matching upstream :shrug:

PR #591, opened 2021-01-26
Add 'docs' to known metadata fields

mthaddon:
Add 'docs' to known metadata fields and warn rather than error on unknown fields in preparation for metadata v2

### DIFF
--- a/charmtools/charms.py
+++ b/charmtools/charms.py
@@ -1055,7 +1055,7 @@ def validate_metadata_keys(charm, linter):
     """
     for key in charm.keys():
         if key not in KNOWN_METADATA_KEYS:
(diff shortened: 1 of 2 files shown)


johnsca (approved):
LGTM, though you should be aware that warnings will still lead to a non-zero exit code (100 rather than 200 for errors; see [`linter.py`](https://github.com/juju/charm-tools/blob/master/charmtools/linter.py)) so any CI checking this which simply consider non-zero a failure will still fail on the warning.  I think it would be fine to consider unknown fields as "info" level lint messages, so that they're still called out in case of typos but are not issues which cause any problem for the charm, necessarily.

johnsca (approved):

PR #592, opened 2021-01-27
Add option to create .charm file

johnsca:
The new pattern used by `charmcraft` is to create a file named `{charm-name}.charm` in the current directory, which is a zip archive of the charm. This adds a `--charm-file` (`-F`) option to also create that file in the same way.

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -1140,8 +1140,8 @@ def main(args=None):
                         help='Same as --log-level=DEBUG')
     parser.add_argument('-c', '--force-color', action="store_true",
                         help="Force raw output (color)")
-    parser.add_argument('--charm-file', '-F', action="store_true",
-                        help="Create a .charm file in the current directory")
+    parser.add_argument('--charm-file', '-F', action='store_true',
+                        help='Create a .charm file in the current directory')
     parser.add_argument('charm', nargs="?", default=".", type=path,
                         help='Source directory for charm layer to build '
                              '(default: .)')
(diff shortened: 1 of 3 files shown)


joedborg (approved):
LGTM

charmtools/build/builder.py
```diff
@@ -1131,6 +1140,8 @@ def main(args=None):
                         help='Same as --log-level=DEBUG')
     parser.add_argument('-c', '--force-color', action="store_true",
                         help="Force raw output (color)")
+    parser.add_argument('--charm-file', '-F', action="store_true",
+                        help="Create a .charm file in the current directory")
```
Cynerva, line 1144: Seems a bit dangerous to drop this in the current directory, e.g. building the easyrsa charm:

```
$ charm-build -F

$ du -h easyrsa.charm
5.1M    easyrsa.charm

$ charm-build -F

$ du -h easyrsa.charm
10M     easyrsa.charm
```

Maybe CHARM_BUILD_DIR would be a better place to put the .charm file?

johnsca: It would be, but `charmcraft` doesn't actually allow you to specify where the file will end up and I wanted to match the behavior of that tool exactly for consistency.

Cynerva: I see. Would it make sense to have the -F flag delete the .charm file prior to build, if it already exists?

johnsca: Added to `DEFAULT_IGNORES` so that we avoid including the `.charm` file in itself.  `charmcraft` needs to do the same.

Cynerva: Sounds good to me. Thanks!

Cynerva (approved):
LGTM

thedac (approved):
Looks good. I was able to build a charm and deploy the zipped version.

PR #598, opened 2021-01-31
Fix reproducible charms issues

ajkavanagh:
The issues fixed in this are:

 * --ignore-lock-file didn't fully ignore the lock file.
 * rebuilding resulted in missing libraries due to purging the wheels.
   Purging was not needed.
 * The wrong fetcher classes were used when rebuilding from lock-files.
   This had the side-effect of not being able to handle multi-layer git
   repositories.  This has been resolved.
 * Fix hard-coded WheelhouseTactic in builder.py
   When doing either the reproducible charms or the wheelhouse overrides,
   the builder used to use the WheelhouseTactic directly.  This breaks any
   code that might override the WheelhouseTactic.  Instead, if an existing
   wheelhouse tactic is in use, use that one's class; if it's not, then
   default back to the base WheelhouseTactic.  May help with bug #585
 * Fix test problems with missing charm_file from Builder()

## Checklist

 - [Y ] Are all your commits [logically] grouped and squashed appropriately?
 - [Y ] Does this patch have code coverage?
 - [Y ] Does your code pass `make test`?

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -171,6 +171,7 @@ class Builder(object):
         self._warned_home = False
         self.lock_items = []
         self.with_locks = {}
+        self.charm_file = False
 
     @property
     def top_layer(self):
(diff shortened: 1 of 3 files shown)


fnordahl (dismissed):
This change resolves issues with building charms consuming interfaces using the layer-index subdir feature

johnsca (dismissed):

johnsca (approved):

johnsca:
Travis is really slow, and the tests pass locally, so I'm moving forward with the merge.

johnsca:
This is out in 2.8.2

PR #600, opened 2021-02-02
Update pinned version of PyYAML

johnsca:
The version of PyYAML pinned in charm-tools is incompatible with the version required by python-libjuju. Updating it seems to not have any noticeable impact on building.


ajkavanagh (approved):
LGTM

PR #601, opened 2021-02-22
Add support for -relation-created hook

johnsca:
Copy the hook template in for -relation-created hooks so that they are handled by the charm.

Also switch to GitHub Actions instead of Travis for tests.


ajkavanagh (approved):
LGTM

Issue #605, opened 2021-04-07
Feature change: charm-build --use-lock-file-branches shouldn't use master branch


ajkavanagh:
## Checklist

 - [X] Confirmed this is an issue with charm-tools, not charmstore-client
 - [X] Provide versions of tools used
 - [X] Described the feature or ways to replicate the issue

## What version am I running?

~I ran the following command: `snap info charm` and got the following ouput:~

charm-tools 2.8.3 from PyPi

I am using: Ubuntu 20.04

## Issue/Feature

When a charm is rebuilt with `--use-lock-file-branches` it *also* uses the `refs/heads/master` branch rather than using the commit hash.  This isn't really what is wanted, I'd hazard a guess (especially as I wrote the feature!)

The idea was that anything that was from a branch would track that branch, but that anything on the master (and I guess 'main') branch would be pinned to the commit hash.  But that's not what happened.

### I expect/expected the following

The master branch would be ignored, and it would select the commit hash.

### What I got

It selects the master branch, rather than the commit hash.

### Proposal

So the principle development branch for git is `master` (or more recently, also `main`).  My proposal is to pick the commit hash if the branch in the `build.lock` file is `refs/heads/master` or `refs/heads/main`.  This will then build the charm according to the stable branches and (if they are not available) commit hashes.

fnordahl:
Fixed by #607

PR #607, opened 2021-04-09
Fixes for reproducible charms (bugs 606, 605, 603)

ajkavanagh:
When the --use-lock-file-branches option was used, several bugs were
discovered in more complicated charms.

* Using the master branch as a tracking branch was the wrong idea.
  Instead, if the master branch is set in the lock file for a layer or
  python module, then the commit hash is used instead.  Use
  --ignore-lock-file to be able to just use master branches. Bug #605
* Detection of branches from a repository is just not possible in
  practice.  So remove the feature and (for compatibility with the build
  from 2.8.3 lockfiles) use the commit hash instead for the branch.
* For python modules use the revision, as detected by th revision-parser
  module, as the locked revision.

Description of change

## Checklist

 - [ ] Are all your commits [logically] grouped and squashed appropriately?


docs/reproducible-charms.rst
```diff
@@ -75,10 +75,13 @@ present:
...
-If the ``--use-lock-file-branches`` is used, then, for VCS items (layers,
-interfaces, and Python modules specified using a VCS string), then the branch
-(if there was one) will be used, rather than the commit version.  This can be
-used to track a branch in a layer or Python module.
+If the ``--use-lock-file-branches`` is used, then, for Python modules specified
+using a VCS string), the branch (if there was one) will be used, rather than
+the commit version.  This can be used to track a branch in a Python module.
+
+Note: as an individual commit be in multiple branches it's not possible to
+detect the branch for a layer.  Thus, this must be overriden manuall in the
```
johnsca, line 83:
```suggestion
detect the branch for a layer.  Thus, this must be overridden manually in the
```

johnsca (approved):
LGTM although ideally this stuff would have test coverage, but I can see how it would be pretty difficult to set up tests for all this.

charmtools/build/builder.py
```diff
@@ -117,8 +117,14 @@ def fetch(self):
                 self.fetched_url = getattr(fetcher, "fetched_url", None)
                 self.vcs = getattr(fetcher, "vcs", None)
             self.revision = fetcher.get_revision(self.directory)
-            self.branch = fetcher.get_branch_for_revision(self.directory,
-                                                          self.revision)
+            # NOTE(ajkavanagh): Due to bug:#606 there isn't a good way to
+            # determine the branch for a revision as there isn't a one-to-one
+            # correspondence.  For the moment, work-around the bug by just
+            # using the revision as the branch so that things do at least
+            # build.
```
johnsca, line 124: I don't actually understand the use-case for "locking" to something mutable like a branch, but if you did want to do this, then you probably want to choose the non-`main`/`master` branch whose head exactly matches the revision, if any, since if you just cloned the repo and got want to be tracking a branch, you should be exactly on that branch at that point. (If my assumption there isn't correct, you could also compute the "closest" branch, i.e. least number of commits delta.) There's still the possibility of ambiguity, but I would expect it to be a lot more rare.

fnordahl (approved):
LGTM

ps: As Cory pointed out, there is a typo in the docs

PR #614, opened 2022-03-25
Fixes for compatibility with Python 3.10

javacruft:
Only pin to earlier versions of some dependencies for older Pythons.

Use collections.abc in preference to deprecated/removed references
to abstract base classes in the collections module itself.


wolsen (approved):
LGTM

ajkavanagh (approved):
LGTM

Issue #615, opened 2022-07-20
validate_series should not raise an error if series is missing


r-f-g:
## Checklist

 - [X] Confirmed this is an issue with charm-tools, not charmstore-client
 - [x] Provide versions of tools used
 - [x] Described the feature or ways to replicate the issue

## What version am I running?

I ran the following command: `snap info charm` and got the following ouput:

```
$ snap info charm
name:      charm
summary:   charm and charm-tools
publisher: Canonical✓
store-url: https://snapcraft.io/charm
contact:   https://discourse.juju.is/c/charming
license:   unset
description: |
  charmstore-client and charm-tools
commands:
  - charm
snap-id:      2Rryoc2ylScfbFl4eQtpntHD9iuZuMvt
tracking:     latest/stable
refresh-date: today at 11:43 CEST
channels:
  latest/stable:    2.8.2               2021-02-01 (609) 119MB classic
  latest/candidate: 2.8.2               2021-02-01 (609) 119MB classic
  latest/beta:      2.8.2               2021-02-01 (609) 119MB classic
  latest/edge:      2.8.3+git-1-736b1ad 2021-02-23 (620) 119MB classic
installed:          2.8.2                          (609) 119MB classic    
```

I am using: 

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04 LTS
Release:        22.04
Codename:       jammy   
```

I also tried using the version installed with pip directly from the master branch ([commit](https://github.com/juju/charm-tools/tree/6f21011580ebf0b4fe7728edcd5010de84baa4b2)).

## Issue/Feature

When I remove `series` from `metadata.yaml`, `charm build` does not work because of these [lines](https://github.com/juju/charm-tools/blob/736b1ade28882290976f4703346d5edc866502c8/charmtools/charms.py#L822-L823). 

In my case, I tried to remove the series from [charm-telegraf](https://git.launchpad.net/charm-telegraf/tree/src/metadata.yaml), which did not work. However, for most reactive openstack charms this will work fine thanks to `layer:openstack-api` which is defined with `series: []`.

### I expect/expected the following

```
$ charm build
...
build: 
build: ---------------------------------------
build:               Build Report
build: ---------------------------------------
build:    metadata.yaml
```

### What I got

```
$ charm build
...
proof: E: missing series: must be a list of series names  
```

ajkavanagh:
[...]

PR #616, opened 2022-07-21
Add warning message if series is in metadata.yaml

r-f-g:
The charmcraft is using bases in charmcraft.yaml instead of series in
metadata.yaml, so the series should not be used anymore.

fixes: #615 

Description of change

## Checklist


ajkavanagh (approved):
LGTM

ajkavanagh:
Based on https://github.com/juju/python-libjuju/pull/671 this seems like a very reasonable update to not actually validate series in the metadata as it is going away.

PR #620, opened 2022-09-07
Add option to build charm with binary wheels

fnordahl:
The reactive framework was born in a time of the now legacy charm store.  The legacy charm store did not have awareness for binary compatibility with various CPU architectures.

In order to enable the use of Python virtual environments, and at the same time support multiple CPU architectures, the charm-tools build action currently uses `pip download` to retrieve the source code of Python dependencies.  On charm install layer-basic will subsequently use pip to build and install the dependencies on the target.

Over the recent years this approach has become increasingly difficult to maintain.  The Python ecosystem has a plethora of build automation and package mangement projects, and each and every Python dependency makes their own choices about what to use.

Adding to that, pip does not really support automatic discovery of build dependencies on download (ref pypa/pip#7863).

Today the legacy charm store has been replaced by charmhub, which does have awareness of CPU architectures.  The Launchpad build automation service has also gained support for charm recipes, which allow developers to build their charms accross a broad range of CPU architectures.

To leverage these capabilities and relieve reactive charm maintainers of the duty of manually hunting down and compiling requirement files for build dependencies, this patch adds support for building charms using binary wheels.

Signed-off-by: Frode Nordahl <frode.nordahl@canonical.com>

### DIFF
--- a/charmtools/build/builder.py
+++ b/charmtools/build/builder.py
@@ -1163,6 +1163,14 @@ def main(args=None):
                         help="Force raw output (color)")
     parser.add_argument('--charm-file', '-F', action='store_true',
                         help='Create a .charm file in the current directory')
+    parser.add_argument('--binary-wheels', action="store_true",
+                        help='Populate the charm wheelhouse with binary '
+                             'wheels that matches the Python version and '
+                             'architecture the charm is built on. Note that '
+                             'to use this option you must ensure that you '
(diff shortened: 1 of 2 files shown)


fnordahl:
The build snap job failure appears to be unrelated to this change.

dshcherb (dismissed):

javacruft (approved):

PR #624, opened 2022-09-15
Archive charm store commands

### LINKED ISSUE #622: Time to archive charm-tools charm store commands / snap build fails on non-x86 architectures due to `bhttp` dependency
Commit 17812f0 fixed a build failure by consuming `bhttp` from another snap, however this snap does not have releases for non-x86 architectures.

The `bhttp` binary is used for the legacy charm store part of charm-tools, specifically in the `promulgate`/`unpromulgate` commands.

Since the charm store is now permanently replaced by charmhub, which is managed by other tools, we should archive these parts of `charm-tools` project.

fnordahl:
The charm store has been decommissioned and replaced by Charmhub.

This tool is still useful for building reactive charms, but store
management should be done with tools such as canonical/charmcraft.

To help reduce the maintenance burden on this tool remove the no
longer used functionality.

Fixes #622


fnordahl:
A version of the snap built on the back of this PR is available in the `latest/edge/archive-charm-store-commands` snap branch.

ajkavanagh:
I guess after this, there should be a major version jump as it's quite the set of changes (in a good way!).

I also need to revamp my PR which fixes some issues around the lock file creation to make it more consistent with multiple branches.  I'll get on to that.

fnordahl:
@ajkavanagh indeed, I also discovered a couple of other issues blocking us from making use of charm-tools from a snap as preferred by the charmcraft reactive plugin.

I've gathered the required changes in a new consolidated pull request #626 and as such I'll close this one.

PR #626, opened 2022-09-16
Archive charm store commands, update snap to core22, when run from snap use python from build environment

fnordahl:
Please review commit by commit.

* Archive charm-store related code and dependencies
  * The charm store has been decommissioned and replaced by Charmhub.
  * This tool is still useful for building reactive charms, but store management should be done with tools such as canonical/charmcraft.
  * To help reduce the maintenance burden on this tool remove the no longer used functionality.

* snap: Update base to core22
* Prune snap path when building wheelhouse
  * When running charm-tools from a snap, the default is to use Python and associated tools from the snap when building a charm's wheelhouse etc.
  * This is normally not what you would want as the Python version in the snap does not necessarily match the Python version of the target system.


fnordahl:
A version of the snap built on the back of this PR is available in the latest/edge/core22 snap branch.

ajkavanagh (dismissed):
LGTM

ajkavanagh (approved):
LGTM - thanks very much for updating the snap build recipe.

PR #627, opened 2022-09-19
Fix non x86 builds of charm-tools itself

fnordahl:
Please review commit by commit.

* setup: Update charm-tools pin set
  * Update the dependency pin set so that dependencies can be
    consumed from versions from the Ubuntu distribution wherever
    possible.

* snap: Ensure build succeeds on non-x86 architectures
  * Update snap recipe so that building charm-tools itself succeeds on non-x86 architectures.
    
  * With reproducibility and maintainability of the snap in mind, make use of Python dependencies from Ubuntu distribution packages wherever possible.
    
  * For any dependencies not available in the Ubuntu distribution, we request a build from source as this will ensure that build-package requirements are uncovered at a early stage in the development process, preferably already in the gate check.
    
  * Drop the PYTHON related build-environment variables as these would mandate installing all build dependencies in the snap, which is not what we want.  Instead reference these variables when needed in the build recipe.


fnordahl:
Binary artifacts available for review in the `latest/edge/fix-non-x86-builds` snap branch.

Build logs per architecture:
[charm_s390x.txt](https://github.com/juju/charm-tools/files/9598685/charm_s390x.txt)
[charm_ppc64el.txt](https://github.com/juju/charm-tools/files/9598686/charm_ppc64el.txt)
[charm_arm64.txt](https://github.com/juju/charm-tools/files/9598687/charm_arm64.txt)
[charm_amd64.txt](https://github.com/juju/charm-tools/files/9598688/charm_amd64.txt)

ajkavanagh (approved):

PR #628, opened 2022-09-19
re-introduce pullsource command without the charmstore fetching capability

addyess:
Description of change

Reintroducing the removed capability to pull sources for layers and interfaces (but without pulling charms from the charmstore)

## Checklist

    - No, then again it never had any before either.


fnordahl (approved):
LGTM, thanks!

PR #633, opened 2022-09-21
Prune GIT environment variables when using Python from buildenv

### LINKED ISSUE #632: snap: git clone may fail when using python from the build environment
```
:: Collecting charms.openstack                                                                                                                    
::   Cloning https://opendev.org/openstack/charms.openstack.git to /root/snap/charmcraft/common/pip-wheel-er66yl_h/charms.openstack               
::   Running command git clone -q https://opendev.org/openstack/charms.openstack.git /root/snap/charmcraft/common/pip-wheel-er66yl_h/charms.openstack                                                                                                                                                
::   warning: You appear to have cloned an empty repository.                                                                                      
::     ERROR: Command errored out with exit status 1:                                                                                             
::      command: /root/snap/charmcraft/common/tmpnokn6imj/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/root/snap/charmcraft/common/pip-wheel-er66yl_h/charms.openstack/setup.py'"'"'; __file__='"'"'/root/snap/charmcraft/common/pip-wheel-er66yl_h/charms.openstack/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /root/snap/charmcraft/common/pip-wheel-er66yl_h/charms.openstack/pip-egg-info                              
::          cwd: /root/snap/charmcraft/common/pip-wheel-er66yl_h/charms.openstack/                                                                
::     Complete output (5 lines):                                                                                                                 
::     Traceback (most recent call last):                                                                                                         
::       File "<string>", line 1, in <module>                                                                                                     
::       File "/usr/lib/python3.8/tokenize.py", line 392, in open                                                                                 
::         buffer = _builtin_open(filename, 'rb')                                                                                                 
::     FileNotFoundError: [Errno 2] No such file or directory: '/root/snap/charmcraft/common/pip-wheel-er66yl_h/charms.openstack/setup.py'        
::     ----------------------------------------                                                                                                   
:: ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.                                
```

It turns out that this is because the snap sets up environment variables assuming `git` will always be called from the snap, but in fact we are using it from the build system.

There are two approaches to this problem
1) We move the snap `git` binary into a folder in the snap where there are no python binaries, so that we can put that folder early in the system `PATH`
2) We extend the `_get_env` helper method to also prune the offending environment variables:
https://github.com/juju/charm-tools/blob/82bf8975293907997fe46986803ad684a13f82ff/charmtools/build/tactics.py#L1066-L1083

The offending environment variables are defined here:
https://github.com/juju/charm-tools/blob/82bf8975293907997fe46986803ad684a13f82ff/helpers/snap-wrappers/charm#L3-L8

fnordahl:
At present we call out to Python provided by the build environment to create a correct wheelhouse.

However, we are not pruning GIT related environment variables set by the snap.  This in turn makes git clone fail when the `git` binary from the build environment attempts to use git resources in the snap, which may be of a different version.

Fixes #632

### DIFF
--- a/charmtools/build/tactics.py
+++ b/charmtools/build/tactics.py
@@ -1073,7 +1073,8 @@ class WheelhouseTactic(ExactMatch, Tactic):
             return os.environ.copy()
 
         env = os.environ.copy()
-        for key in ('PREFIX', 'PYTHONHOME', 'PYTHONPATH'):
+        for key in ('PREFIX', 'PYTHONHOME', 'PYTHONPATH',
+                    'GIT_TEMPLATE_DIR', 'GIT_EXEC_PATH'):
             if key in env:
                 del(env[key])
         env['PATH'] = ':'.join([


ajkavanagh (commented):
LGTM

ajkavanagh:
So this change will assume that git is installed on the host system, bypassing the version in the snap.  Git is *almost* always installed so it *should* be okay?

fnordahl:
> So this change will assume that git is installed on the host system, bypassing the version in the snap. Git is _almost_ always installed so it _should_ be okay?

Essentially yes, this goes in hand with our need for using Python from the build environment and not from the snap. Commit 2482f0f104d31f6f3f50e5645d0705bb9b738a72 changed the default behavior, and at the same time added a option to use the old behavior.

So if anyone would be harmed by this change they have a way to override it.

ajkavanagh (approved):
LGTM

PR #637, opened 2022-09-29
Fix version output to be compatbile with sibling projects

### LINKED ISSUE #636: charm version compatibility output is unfortunately not compatible with charmcraft 2.0
The [code in charmcraft 2.0](https://github.com/canonical/charmcraft/blob/dbf82a0fbaa6633248bb213d3943603c1e7c50a6/charmcraft/reactive_plugin.py#L64-L65) looks like this:
```python
            output = self._execute("charm version").strip()
            _, tools_version = output.split("\n")
```

And our current compatibility code look like this:
https://github.com/juju/charm-tools/blob/e1c7ee0b994e12660257e4021bef487d6dbf01a9/charmtools/version.py#L70-L75

Due to charmcraft's use of [strip](https://docs.python.org/3/library/stdtypes.html#str.strip) as opposed to [rstrip](https://docs.python.org/3/library/stdtypes.html#str.rstrip), this does not work.

Example:
```
>>> s = subprocess.run(["/bin/bash", "-c", "charm version"], check=True, capture_output=True, universal_newlines=True)
>>> s
CompletedProcess(args=['/bin/bash', '-c', 'charm version'], returncode=0, stdout='\ncharm-tools 3.0.0+snap-661+git-1-b7025b2\n', stderr='')
>>> s.stdout.strip()
'charm-tools 3.0.0+snap-661+git-1-b7025b2'
```

Now if we fudge the output by adding something on the first line:
```
>>> s = subprocess.run(["/bin/bash", "-c", "charm version"], check=True, capture_output=True, universal_newlines=True)
>>> s
CompletedProcess(args=['/bin/bash', '-c', 'charm version'], returncode=0, stdout='-\ncharm-tools 3.0.0+snap-x1+git-3-e1c7ee0\n', stderr='')
>>> s.stdout.strip()
'-\ncharm-tools 3.0.0+snap-x1+git-3-e1c7ee0'
>>> _, tools_version = s.stdout.strip().split('\n')
>>> tools_version
'charm-tools 3.0.0+snap-x1+git-3-e1c7ee0'
```

fnordahl:
Due to how some other projects parse our version string the current format may break people.

Fixes #636

### DIFF
--- a/charmtools/version.py
+++ b/charmtools/version.py
@@ -70,7 +70,7 @@ def main():
         # ensure compatibility with anyone parsing this output expecting the
         # charm-tools version on the second line ref:
         # https://github.com/canonical/charmcraft/blob/dbf82a0fb/charmcraft/reactive_plugin.py#L65
-        print("")
+        print("-")
         print("charm-tools {}".format(
             format_version(cached_charm_tools_version(), args.format)))
 


freyes (approved):
LGTM, all the reasoning described in #636 looks correct to me.

Issue #646, opened 2022-10-10
No control over virtualenv used to build wheels


fnordahl:
Some python packages may require changes to the virtualenv used to build wheels. There is currently no way of doing that.

One example is Jinja2 when building on Focal (ref pallets/jinja#1496):
https://github.com/juju/charm-tools/blob/977ec1ff795f5a8692f8061fc16aa01adf9478cf/.github/workflows/tests.yml#L57-L66

The virtualenv is created by charm-tools here:
https://github.com/juju/charm-tools/blob/977ec1ff795f5a8692f8061fc16aa01adf9478cf/charmtools/build/tactics.py#L1241-L1244

facundobatista:
+1 to this issue.

When `charmcraft` packs a charm using the `charm` plugin, it builds a virtualenv and then allows the user to install packages there before actually dealing the charm requirements. This way, the user can upgrade `setuptools` and `pip`, a step that is needed for later proper installation of Jinja2 (the Focal default versions of `setuptools` and `pip` do not handle that correctly).

But in the case of `charmcraft` packing a charm using the `reactive` plugin, it will end up calling `charm`, which is the one responsible of the virtualenv creation and installation of charm requirements. As everything is handled in the same `charm` tool process, there is no way `charmcraft` can get in the middle and update these packages "from outside".

So it would be awesome if the `charm` tool handleds this. Thank you!!

hloeung:
+1 here too. With charm tools 2.x, we can override python packages in `wheelhouse.txt`. This breaks with 3.x (at least with `3.0.4`).

Downgrading to `2.x/stable` either in charmcraft.yaml with `build-snaps: [charm/2.x/stable]` as mentioned in https://github.com/juju-solutions/layer-basic/issues/209 fixes this for me.

lathiat:
Does this change fix this:
https://discourse.charmhub.io/t/install-or-update-python-packages-before-packing-a-charm/5158

See also: https://github.com/canonical/charmcraft/issues/551

fnordahl:
> Does this change fix this: https://discourse.charmhub.io/t/install-or-update-python-packages-before-packing-a-charm/5158
> 
> See also: [canonical/charmcraft#551](https://github.com/canonical/charmcraft/issues/551)

No, when building with the reactive plugin, charm-tools is in charge of the virtualenv, the fix in charmcraft only affects building with the charm plugin for operator framework charms. See also https://github.com/juju/charm-tools/issues/646#issuecomment-1273451182

PR #649, opened 2022-11-08
Add a deprecation/status message to the README

ajkavanagh:
This is to effectively disuade people to use it and get them to look elsewhere for writing new charms (i.e. point them to the Operator Framework).  Also add details of which track to use for which version of charms.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,9 +13,82 @@ home page for more information.
 
 https://jujucharms.com/
 
+STATUS
+======
+
+charm-tools is now largely in *maintenance mode*, as the charms.reactive method
+of building charms is deprecated in favour of using the `operator
+framework <https://juju.is/docs/sdk>`_.  This means that, generally, only bug
+fixes or features required to make charm-tools work better with charmcraft will
+be considered.
+
+charm-tools should only be used to maintain existing charms.reactive charms,
+and not to start *new* charms.  The `operator framework
+<https://juju.is/docs/sdk>`_ should be used for new charms.
+
+Also, it is increasingly more difficult to get a single Python code base to
+build source charms that will work across Xenial to Jammy+ (e.g. Python 3.5 to
+Python 3.10) due to Python versions < 3.7 being EOL. There are a number of
+strategies to deal with that, and it generally involves pinning python modules
+for various different versions of Python when building the charm.  As the
+`charmhub <https://charmhub.io/>`_ supports architecture-specific builds of
+charms, charm-tools also offers binary charms, which are charms with
+pre-compiled binary wheels that do not need build environments on the target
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -13,9 +13,82 @@ home page for more information.
...
+    type: charm
+
+    parts:
+      charm:
+        source: src/
+        plugin: reactive
+        build-snaps:
+          - charm/3.x/stable
+        build-environment:
+          - CHARM_INTERFACES_DIR: /root/project/interfaces/
```
fnordahl, line 54: You could also use `$CRAFT_PROJECT_DIR` here

README.rst
```diff
@@ -13,9 +13,82 @@ home page for more information.
...
+    bases:
+      - build-on:
+          - name: ubuntu
+            channel: "22.04"
+            architectures:
+              - amd64
+        run-on:
+          - name: ubuntu
+            channel: "22.04"
+            architectures: [amd64, s390x, ppc64el, arm64]
```
fnordahl, line 66: This specification is incorrect, and will not work to build multiple charm artifacts using Launchpad. Ref: https://bugs.launchpad.net/launchpad/+bug/1967319

README.rst
```diff
@@ -13,9 +13,82 @@ home page for more information.
...
+it's building on the 22.04 base (jammy) which is Python 3.10.
+
+Snap Tracks/Version
+-------------------
+
+Due to the complexity of building reactive charms across multiple Python
+versions, the tool has been split into two tracks:
+
+1. For charms targetting Ubuntu focal (20.04) or earlier: 2.x
+2. For charms targetting Ubuntu jammy (22.04) or later: 3.x
```
fnordahl, line 81: This is a gross oversimplification of the problem at hand, and I do not think this will help guide people into making good decisions.

Why not instead cover the fact that charm-tools 2.x is hard coded to build using Python 3.6, which is equivalent of using a Ubuntu 18.04 (Bionic) base, and that charm-tools 3.x uses the Python from your build environment?

The Focal setuptools problem as revealed by the upstream Jinja2 [issue](https://github.com/pallets/jinja/issues/1496) is of course worth noting, and patches are of course welcome to actually fix the root of the issue as outlined in #646.

fnordahl (changes requested):
This is a welcome addition to the guidance to use other tools for new charms that was added to the documentation when we created 3.x ref https://github.com/juju/charm-tools/commit/679a9f05cfaf2cbe47a5949475600bbe79ec469c.

The text does include a couple of things that needs fixing though, see in-line.

ajkavanagh:
Ping - just seeing if this is still needed.

PR #650, opened 2022-11-12
Add option to upgrade core deps in build venv

fnordahl:
Some packages may require newer versions of core dependencies for successful execution of the `pip download` or `pip wheel` commands. Most notable example is Jinja2 and MarkupSafe when installed on Focal ref pallets/jinja#1496.

At this stage of the process the input from the charm wheelhouse.txt is not sufficient.

Add an option to upgrade pip and setuptools in the build venv to deal with these situations.

Fixes: #646
Signed-off-by: Frode Nordahl <frode.nordahl@canonical.com>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,17 +54,6 @@ jobs:
       with:
         repository: juju-solutions/layer-basic
 
-    - name: Fixup wheelhouse
-      run: |
-        set -euxo pipefail
-        cat << EOF | tee -a tests/charm-minimal/wheelhouse.txt
-        # https://github.com/pallets/jinja/issues/1496
-        #
-        # We ought to teach charm-tools to seed the virtualenv used to build
-        # wheels with newer versions of pip and setuptools.
-        Jinja2<3
-        EOF
-
     - name: Download built charm snap
       uses: actions/download-artifact@v3
       with:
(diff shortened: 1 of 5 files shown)


lathiat:
Does this fix the same issue, but with a charmcraft.yaml change rather than a config option?
https://discourse.charmhub.io/t/install-or-update-python-packages-before-packing-a-charm/5158

Not clear to me if the proposed fix here works in some cases the above wouldn't. But the above fix went in only a couple weeks before this bug/fix so not sure if it was just missed.

(Originally from: https://github.com/canonical/charmcraft/issues/551)

fnordahl:
> Does this fix the same issue, but with a charmcraft.yaml change rather than a config option? https://discourse.charmhub.io/t/install-or-update-python-packages-before-packing-a-charm/5158
> 
> Not clear to me if the proposed fix here works in some cases the above wouldn't. But the above fix went in only a couple weeks before this bug/fix so not sure if it was just missed.
> 
> (Originally from: [canonical/charmcraft#551](https://github.com/canonical/charmcraft/issues/551))

The charmcraft fix is only usable by the charm plugin for operator framework charms. When building with the reactive plugin charm-tools is in charge of the virtualenv.

Once this PR is merged, you would be able to control this through the charmcraft.yaml like this:
```yaml
type: charm

parts:
  charm:
    source: src/
    plugin: reactive
    reactive-charm-build-arguments:
      - --binary-wheels-from-source
      - -v
      - --upgrade-buildvenv-core-deps
```

ajkavanagh (approved):

PR #651, opened 2023-02-07
Fix a couple of pip requirement lines

ajkavanagh:
A change in the strictness of how pip requirements lines are parsed means that the lib needs a tactical fix to enable testing to move forwards.  A more complete fix is to remove the upper bounds and the main development branch.


javacruft:
The integration test is failing due to an issue using charmcraft rather than something related to this change to bypassing branch protections to unblock the wider world with this fix.

Issue #652, opened 2023-03-09
version 3.0.4 (PyPi) fails building octavia due to setuptools 67+ being very strict on PEP 440


ajkavanagh:
## Checklist

 - [X] Confirmed this is an issue with charm-tools, not charmstore-client
 - [X] Provide versions of tools used
 - [X] Described the feature or ways to replicate the issue

## What version am I running?

I'm using 3.0.4 from PyPi inside charmcraft.

I am using: Ubuntu 22.04

Versions of relevant modules within charm-build run:

```
:: charmtools.build.tactics: Packages in buildvenv:                                                                                                                                                                    
:: Package    Version                                                                                                                                                                                                  
:: ---------- -------                                                                                                                                                                                                  
:: pip        23.0.1                                                                                                                                                                                                   
:: setuptools 67.4.0                                                                                                                                                                                                   
:: wheel      0.38.4 
```

## Issue/Feature

Charm build fails with:

```
:: Collecting kombu==5.2.4                                                                                                                                                                                             
::   Downloading kombu-5.2.4.tar.gz (423 kB)                                                                                                                                                                           
::      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 423.4/423.4 kB 4.2 MB/s eta 0:00:00                                                                                                                                    
::   Preparing metadata (setup.py): started                                                                                                                                                                            
::   Preparing metadata (setup.py): finished with status 'error'                                                                                                                                                       
::   error: subprocess-exited-with-error                                                                                                                                                                               
::                                                                                                                                                                                                                     
::   × python setup.py egg_info did not run successfully.                                                                                                                                                              
::   │ exit code: 1                                                                                                                                                                                                    
::   ╰─> [5 lines of output]                                                                                                                                                                                           
::       /root/snap/charmcraft/common/tmp7pqi369d/lib/python3.10/site-packages/setuptools/config/setupcfg.py:520: SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.   
::         warnings.warn(msg, warning_class)                                                                                                                                                                           
::       error in kombu setup command: 'tests_require' must be a string or list of strings containing valid project/version requirement specifiers; Expected end or semicolon (after name and no valid version specifier)                                                                                                                                                                                                                      
::           pytz>dev                                                                                                                                                                                                  
::               ^                                                                                                                                                                                                     
::       [end of output]
```

[...]

PR #653, opened 2023-03-09
Pin pip<23.1 and setuptools<67 to allow PEP-440 non compliance

ajkavanagh:
PEP-440 is more strict about requirement lines.  As an example:

    pytz>dev

is non-compliant, but is in 5.2.4 of kombu's requirements.  This breaks building wheelhouses from source (such as the octavia charm).  This patch pins pip and setuptools to the latest versions that will defintely still allow PEP-440 non-compliant packages to install.

The option --upgrade-buildvenv-core-deps can be used to override this and will install the latest versions of pip and setuptools available.

Fixes-Bug: #652

Description of change

## Checklist


ajkavanagh:
The integration test failure is unrelated to the change; sadly it looks like a networking issue within charmcraft.

PR #656, opened 2023-04-05
PyYAML>=5.0,<7.0

freyes:
ops>-2.1.0 depends on PyYAML-6.x

Description of change

## Checklist

 - [x] Are all your commits [logically] grouped and squashed appropriately?
 - [ ] Does this patch have code coverage?
 - [ ] Does your code pass `make test`?

PR #659, opened 2023-07-20
setup: Update charm-tools pin set

coreycb:
pyyaml 5.4.0 and 5.4.1 are broken with cython 3
https://github.com/yaml/pyyaml/issues/724

jsonschema 4.18.0 depends on Rust (via rpds-py)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,6 +49,10 @@ jobs:
         sudo lxd init --auto
         # This is a throw-away CI environment, do not do this at home
         sudo chmod 666 /var/snap/lxd/common/lxd/unix.socket
+        # Fixup iptables forwarding issues from LXD containers with a flush and
+        # re-create of rules.
+        sudo iptables -F FORWARD
+        sudo iptables -P FORWARD ACCEPT
     - name: Checkout layer-basic
       uses: actions/checkout@v2
       with:
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
         exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
     install_requires=[
         'cheetah3>=3.0.0,<4.0',
-        'pyyaml>=5.0,<6.0',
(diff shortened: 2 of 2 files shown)


ajkavanagh:
This is similar to #656 - could we please consolidate them?  So we also need to pin `Cython<3.0.0` as that's also broken with PyYaml 6.0.1 (at the moment).

ajkavanagh (changes requested):
Do we also need to pin Cython?

ajkavanagh (approved):

freyes:
we need a 3.0.7 release in pypi with this fix

markbeierl:
This change should be revisited now that 6.0.1 was released to fix the problem.  I'm not sure, but it looks like the result of         'pyyaml>=5.0,!=5.4.0,!=5.4.1,<6.0' is 5.3.1, which has known CVEs.

```
pip install 'pyyaml>=5.0,!=5.4.0,!=5.4.1,<6.0'
Collecting pyyaml!=5.4.0,!=5.4.1,<6.0,>=5.0
  Using cached PyYAML-5.3.1-cp311-cp311-linux_x86_64.whl
Installing collected packages: pyyaml
Successfully installed pyyaml-5.3.1
```

freyes:
On Wed, 2023-08-09 at 12:58 -0700, Mark Beierl wrote:
> This change should be revisited now that 6.0.1 was released to fix the problem.  I'm not sure, but
> it looks like the result of         'pyyaml>=5.0,!=5.4.0,!=5.4.1,<6.0' is 5.3.1, which has known
> CVEs.
> pip install 'pyyaml>=5.0,!=5.4.0,!=5.4.1,<6.0'
> Collecting pyyaml!=5.4.0,!=5.4.1,<6.0,>=5.0
>   Using cached PyYAML-5.3.1-cp311-cp311-linux_x86_64.whl
> Installing collected packages: pyyaml
> Successfully installed pyyaml-5.3.1

I think we could do:

pyyaml>=5.0,!=5.4.0,!=5.4.1,!=6.0,<7.0

freyes:
https://github.com/juju/charm-tools/pull/663

PR #671, opened 2024-01-05
setup: pin macaroonbakery<1.3.3 for python3.6

lathiat:
macaroonbakery in v1.3.4 added a constraint of protobuf>=3.20.0[0] which makes it incompatible with python 3.6 and v1.3.3 was released in a broken state[1].

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,13 @@ setup(
         'pyyaml>=5.0,!=5.4.0,!=5.4.1,<6.0',
         'requests>=2.0.0,<3.0.0',
         'libcharmstore',
+        # macaroonbakery is pulled in by libcharmstore->theblues, v1.3.4 added
+        # a constraint of protobuf>=3.20.0[0] which makes it incompatible with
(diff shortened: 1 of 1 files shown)


lathiat:
I'm still having trouble with octavia/ussuri failing to build due to macaroonbakery/protobuf (https://review.opendev.org/c/openstack/charm-octavia/+/902746).

It's only broken for Python3.6 (Bionic) which only effects ussuri, victoria is fine since it only needs Python3.8 (Focal).

The specific issue is charm-tools < 3.0.0 -> libcharmstore -> theblues -> macaroonbakery

charm-tools 3.x actually installs and works fine on python3.6 (and works to build the charm) because libcharmstore was dropped, but the stable charms currently tend to pin charm-tools to <3.0.0 (or == 2.8.x)

coreycb (approved):
This looks good to me. I can't merge it though.

freyes (approved):
LGTM. This is the approach we have taken elsewhere.

PR #676, opened 2024-07-31
Update github workflow to include a noble build on charmcraft-2.x

addyess:
Update github action workflows which are marked as [deprecated](https://github.com/juju/charm-tools/actions/runs/10185563744). 

While we're at it-- confirm that noble builds work on these tools

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,9 +16,9 @@ jobs:
           - '3.12'  # noble
     steps:
     - name: Check out code
-      uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -107,6 +107,9 @@ jobs:
           - name: ubuntu
             channel: "22.04"
             architectures: [amd64]
+          - name: ubuntu
+            channel: "24.04"
+            architectures: [amd64]
```
addyess, line 112: adds a noble base to try out.

PR #685, opened 2024-12-02
add support for python 3.13

branchvincent:
This PR contains a couple of fixes for python 3.13:

* `cheetah3` has been [renamed to `ct3`](https://github.com/CheetahTemplate3/cheetah3/commit/673259b2d139b4ea970b1c2da12607b7ac39cbec), with the former no longer being updated and broken on 3.13: https://github.com/CheetahTemplate3/cheetah3/issues/63
* Invalid escape sequences now raise a `SyntaxWarning` on 3.12+: https://docs.python.org/3/whatsnew/3.12.html#other-language-changes
* While here, migrate mostly to [`pyproject.toml`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) (the modern replacement for `setup.py` favoring static metadata). Now only the dynamic `version` remains in `setup.py`.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,63 @@
+[project]
+name = "charm-tools"
+description = "Tools for building and maintaining Juju charms"
+readme = "README.rst"
+maintainers = [{ name = "Cory Johns", email = "johnsca@gmail.com" }]
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Programming Language :: Python",
+]
+dependencies = [
+    "blessings<2.0",
+    "colander<1.9",
+    "ct3>=3.0.0,<4.0",
+    "dict2colander==0.2",
+    "jsonschema<4.18.0",
+    "jujubundlelib<0.6",
+    "keyring<24",
+    "otherstuf<=1.1.0",
+    "path<17",
+    "pathspec<0.11;python_version >= '3.7'",
+    "pathspec<=0.3.4;python_version < '3.7'",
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+python = "charmtools.templates.python_services:PythonServicesCharmTemplate"
+python-basic = "charmtools.templates.python:PythonCharmTemplate"
+reactive-bash = "charmtools.templates.reactive_bash:ReactiveBashCharmTemplate"
+reactive-python = "charmtools.templates.reactive_python:ReactivePythonCharmTemplate"
+
+[tool.setuptools.packages.find]
+exclude = ["*tests*"]
+
+[build-system]
+requires = ["setuptools>=68"]
```
freyes, line 62: this is an aspect that makes be concerned, how compatible this is with older versions of python, do you know what versions can we support with pyproject.toml ?

freyes (changes requested):
Hi, thanks for this PR, this is something we must definitively want to have, I just left one comment inline, which prompts me to think that maybe we would need to cut a new major release (charm-tools 4.x ) to drop older versions of python

freyes (changes requested):
Add 3.13 to the testing matrix please, - https://github.com/juju/charm-tools/blob/master/.github/workflows/tests.yml#L16 - also we need a maintainer to allow this PR to run the CI

pyproject.toml
branchvincent: Sounds good, added a `requires-python` to make that support explicit

PR #696, opened 2026-04-28
feat: reactive on 26.04

sabaini:
Reactive plugin support for 26.04


sabaini:
charmcraft currently blocks reactive charms on 26.04 -- i have exercised this PR with a local charmcraft which has the block removed
Example run: https://gist.github.com/sabaini/2796b805877b65145483a8986774c215

sabaini:
also cf. https://github.com/canonical/charmcraft/pull/2678

snap/snapcraft.yaml
```diff
@@ -78,8 +78,15 @@ parts:
...
       pip install --upgrade 'setuptools<82'
+      # Most dependencies are built from source for the snap, but modern
+      # virtualenv and its pure-Python dependencies use hatchling. Building
+      # those sdists inside the staged core22 environment can import the
+      # staged python3-pathspec, which is too old for hatchling. Use wheels for
+      # that pure-Python dependency set while keeping source builds for the
+      # rest of the dependency set.
       pip install \
           --no-binary :all: \
+          --only-binary virtualenv,distlib,filelock,platformdirs,python-discovery \
```
freyes: this will break (most likely) some of the not so common architectures we support, e.g. ppc64el and s390x

freyes (changes requested):

snap/snapcraft.yaml
sabaini: Ok I've now switched to core24 so we get a newer python3-pathspec, this otherwise breaks hatchling

Also had to introduce Cython min. version as this would otherwise break

sabaini:
@freyes I believe the charmcraft changes have landed now in https://github.com/canonical/charmcraft/pull/2678 so tests should have a better chance of passing

